PR #37, opened 2026-08-12
feat(theme): probe terminal colors via OSC and derive theme automatically

r17x:
## Summary

~~Bohay now **auto-adapts to your terminal's color scheme** out of the box. Instead of shipping a fixed default palette, it probes your terminal's actual foreground, background, and ANSI colors at startup, then derives the entire UI theme from them.~~

~~- Looks native in any terminal — dark, light, or custom~~
~~- No manual theme selection needed; just works~~
~~- Falls back to `quattro-rally` when probing isn't possible (piped stdin, non-Unix)~~
~~- Colors are bridged to the server process so client/server sessions stay consistent~~

Adds a **"terminal"** theme that probes the host terminal's actual colors at startup via OSC escape sequences, then derives the entire UI from them.

- Opt-in: set `"theme": "terminal"` in config — zero cost when not selected
- Probe is gated on config, exits early once all 18 responses arrive (~10ms)
- Non-OSC bytes typed during probe are stuffed back to stdin via TIOCSTI
- Colors bridge to the server process via `BOHAY_TERMINAL_COLORS` env var
- Falls back to `quattro-rally` when probing is unavailable (no tty, Windows, `--remote`)
- All color math unified to a single `blend_rgb(a, b, t)` primitive
- Surface/border derivation blends toward `palette[8]` (bright black) so chrome tint matches the host scheme

## Call Graph

```
main()
├── run_local()
│   └── config::load().theme == "terminal"?
│       └── probe_terminal()
│           ├── OnceLock guard
│           ├── enable_raw_mode()
│           ├── write OSC 10/11/4 queries to stdout
│           ├── libc::poll + stdin read loop (200ms deadline)
│           │   └── count_osc_terminators() → early exit at 18
│           ├── disable_raw_mode()
│           ├── extract_non_osc_bytes(buf) → stuff_stdin(leftover)
│           ├── parse_osc_responses(buf)
│           │   ├── parse_rgb_value() ×18
│           │   └── default_ansi_palette(fg, bg)
│           ├── encode_colors() → set BOHAY_TERMINAL_COLORS env var
│           └── PROBED.set(colors)
│
└── autodetect_and_attach()
    └── config::load().theme == "terminal"?
        └── probe_terminal()
            └── ... → spawn_server() inherits env var
                       └── server process:
                           by_name("terminal")
                           └── probed_colors()
                               ├── PROBED.get()
                               └── decode_colors(env) → PROBED.set()
                           └── Theme::from_terminal(colors)
                               ├── luminance() → dark/light
                               ~~└── from_terminal_base() + struct spread~~
                                   ~~└── map_channels / blend_rgb / pal~~
                               └── from_terminal_base() + blend_rgb / pal
```

## Summary by CodeRabbit

* **New Features**
  * Added a terminal theme that automatically matches detected terminal colors.
  * Added terminal color detection with support for dark and light palettes.
  * Preserved keyboard input while terminal color detection is in progress.
  * Added connection-time theme negotiation for remote sessions.

* **Bug Fixes**
  * Improved focus recovery by treating regained focus as a resize event.
  * Improved handling of terminal input events during initialization.


RizRiyz:
Thanks @r17x nice work! The OSC parsing is solid and deriving the palette from fg/bg plus the 16 ANSI colours is a good approach. 

A few things

**CI is red** 
`cargo fmt --all` fixes one diff at `theme.rs:1147`. Two tests fail on the changed default (`config::tests::defaults_and_roundtrip` and `app::tests::settings_modal_interactions`). Note fmt runs before clippy in that job, so clippy never ran, worth checking locally with `-D warnings`.

**The probe adds ~200 ms to every launch, and everyone pays it**
This is the main thing. `probe_terminal()` is the first line of both `run_local()` and `autodetect_and_attach()`, before config is read, so it runs whether or not the terminal theme is in use.

Two changes needed:

1. Read config first and only probe when the theme is actually `terminal`, so people who don't use it pay nothing.
2. Give the loop an early exit. Right now there isn't one:

```rust
let ret = unsafe { libc::poll(fds.as_mut_ptr(), 1, remaining.as_millis() as i32) };
if ret <= 0 { break; }
```

The terminal replies in a few ms, then it polls again with the remaining ~195 ms and waits out the timeout. Bailing once 18 responses have parsed, or counting `BEL`/`ST`, should get it to ~10 ms.


Also check [CONTRIBUTING](../blob/main/CONTRIBUTING.md#guidelines) that mention to keep blocking I/O off the startup and event paths, and 200 ms is very visible on something people launch all day.


**It swallows stdin** 
Reading stdin in raw mode before `ratatui::init()` means anything typed during startup goes into the probe buffer and is lost. Could you push non-OSC bytes back?

**Make it a theme people pick rather than the default** 
It behaves differently from the other palettes in two cases: over `--remote` the env var does not cross the SSH boundary so it falls back, and with several clients attached the first one's colours win for everyone. As an opt-in that's fine, as the default it will generate confused issues. It also means both failing tests pass untouched.

Smaller stuff:

- The registry test now skips `terminal` in the swatch check. You could assert against the fallback palette instead, so the entry stays covered
- The `.gitignore` nix lines look unrelated, happy to take them either way.
- A one-line comment on why `set_var` is safe here would help later readers.

The hard parts look good. Mainly green CI, the early exit, and opt-in rather than default.

r17x:
> The .gitignore nix lines look unrelated, happy to take them either way.

@RizRiyz I'm working with Nix environment, When I build for local development, I don't wan to see `result/` based on the nix build command. So, It normal to ignore `result/`

RizRiyz:
Thanks for terminal theme @r17x

I tested the feature on Ghostty and pushed a follow-up hardening commit to the PR. The main improvements are:

  - Terminal color probing now runs only when the `terminal` theme is selected
  - Other themes perform no probe and have no probe-related startup cost
  - Probing is handled by the foreground client, including remote attachments
  - Keyboard input received during probing is preserved safely
  - Unsupported terminals fall back after a bounded 50 ms timeout
  - Failed or disabled color reporting uses the terminal's native ANSI palette instead of Quattro Rally
  - The protocol handshake explicitly negotiates whether probing is needed
  - Terminal is placed at the bottom of the theme picker as the newest theme
  - The PR has been updated with the latest `main`

Performance testing found no render or ongoing memory regression. A successful Ghostty color probe adds approximately 1.2 ms during attachment

PR #61, opened 2026-08-20
fix(windows): AltGr must type its character, not a Ctrl chord

### LINKED ISSUE #103: Windows: incorrect keyboard layout / special characters in terminal panes
## Description

On Windows, keyboard input inside Luvus terminal panes appears to use an incorrect keyboard layout or incorrectly handles some key combinations.

I am using a Spanish keyboard layout.

PowerShell works correctly when opened normally outside Luvus, but the same characters are interpreted incorrectly when PowerShell is running inside a Luvus terminal pane.

This particularly affects special characters and AltGr combinations.

## Examples

Characters such as:

```text
@ \ | { } [ ] ~ € ñ á
```

work correctly in a normal PowerShell window, but some of them produce incorrect characters inside Luvus.

Paste also appears to be affected in some cases.

For example, pasting Windows commands/paths can result in corrupted input. I have seen commands similar to:

.\.venv\Scripts\python.exe .\youtube_folder_uploader.py --folder "E:\Videos\Pendientes"

become malformed inside the Luvus terminal.

Expected behavior

Luvus terminal panes should respect the active Windows keyboard layout and produce exactly the same characters as a normal PowerShell / Windows Terminal session.

Actual behavior

Special characters and/or AltGr key combinations can produce different characters inside Luvus.

The same keyboard input works correctly in PowerShell outside Luvus.

Environment
OS: Windows
Shell: PowerShell
Keyboard layout: Spanish
Luvus: current version
Additional context

This looks potentially related to the terminal input handling rather than PowerShell itself.

Luvus appears to capture keyboard events through crossterm and then encode/forward them to the PTY. Paste input also appears to be handled by Luvus before being sent to the focused pane.

Alex-Blanes:
## The problem

The Windows console reports AltGr as `CONTROL | ALT` — the layout driver presses left-Ctrl together with right-Alt — so every place that asked `modifiers.contains(CONTROL)` swallowed the AltGr characters.

On a Spanish keyboard those are `\ @ # [ ] { } | ~ €`:

- in a pane, `\` was sent as `0x1c` and `[` as a bare `ESC`; `# { } | ~` were dropped outright (`encode_key` returned `None`)
- the global search (`search.rs`) and the tab switcher (`switcher.rs`) ignored them the same way

The practical result: you cannot type a Windows path inside luvus.

## The fix

`keys::is_ctrl_chord()` — Ctrl, but not Ctrl+Alt on Windows — and all three sites go through it. `encode_key` also stops prefixing `ESC` for the Alt half of an AltGr press.

Windows-only by design: X11 and macOS terminals deliver an AltGr character with **no modifiers at all**, so there `Ctrl+Alt+<key>` is a real chord and keeps working as before. The test asserts both halves.

## Summary by CodeRabbit

* **New Features**
  * Added fuzzy search across sessions, workspaces, tabs, panes, agents, files, terminal output, and remote sessions.
  * Added search scopes, recommendations, paging, recent-file results, and improved activation handling.
  * Added fuzzy matching when switching between workspaces.
  * Added configurable Copy Mode and expanded function-key shortcuts.

* **Bug Fixes**
  * Improved Windows AltGr and modifier handling.
  * Preserved pasted paths, quotes, Unicode characters, and command text across platforms.
  * Improved mouse navigation, scrolling, menus, and copy selection behavior.
  * Improved backend event handling and pending request completion.

## Updated for current main

- rebased onto the latest `main` input encoder
- preserves application cursor mode and non-Windows Ctrl+Alt behavior
- covers `€`, `ñ`, and `á` alongside the original AltGr characters
- verifies pasted Windows paths, quotes, and Unicode at both input boundaries
- runs the focused AltGr regression tests in Windows CI

Fixes #103

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,6 +89,7 @@ jobs:
           cargo test --locked ipc::transport::windows_
           cargo test --locked app::backend::tests
           cargo test --locked terminal::backend::tests
+          cargo test --locked altgr
       - name: Validate published protocol fixtures
         shell: pwsh
         run: |
--- a/src/app/input.rs
+++ b/src/app/input.rs
@@ -3186,11 +3186,10 @@ mod tests {
     fn altgr_types_its_character_instead_of_a_control_byte() {
         let nl = b"\x1b\r";
         let altgr = KeyModifiers::CONTROL | KeyModifiers::ALT;
-        let enc = |c: char, m: KeyModifiers| {
-            encode_key(&KeyEvent::new(KeyCode::Char(c), m), nl, false)
-        };
+        let enc =
+            |c: char, m: KeyModifiers| encode_key(&KeyEvent::new(KeyCode::Char(c), m), nl, false);
         if cfg!(windows) {
-            for c in ['\\', '@', '#', '[', ']', '{', '}', '|', '~'] {
+            for c in ['\\', '@', '#', '[', ']', '{', '}', '|', '~', '€'] {
                 assert_eq!(
                     enc(c, altgr),
                     Some(c.to_string().into_bytes()),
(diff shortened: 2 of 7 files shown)


RizRiyz:
@Alex-Blanes can you rebase and fix this PR, its related to #103. I am happy to take a look and approve this fix.

PR #62, opened 2026-08-20
feat(picker): paste a path to jump there, instead of leaking it to the pane

Alex-Blanes:
## The problem

`paste_into_modal` (`app/input.rs`) listed the rename prompts but neither the folder picker nor the global search, even though both capture the keyboard in `handle_key`. Pasting with either of them open sent the text **to the pane behind the modal**: a pasted path ended up in whatever shell or agent happened to be focused.

And the picker only navigates with `j/k/h/l`, so there was no way to reach a deep folder or another drive without walking the tree row by row.

## The fix

- The global search receives the paste as its query.
- The picker has no text field, so a pasted path **is** the navigation: it jumps there. Surrounding whitespace and quotes are stripped (what a file manager actually puts on the clipboard), a pasted file lands on its parent folder, and anything that does not resolve leaves the browsed folder alone and says so in the modal instead of silently doing nothing.
- The picker's "new folder" sub-mode still receives the paste as text.

Follows #61: with AltGr fixed you can *type* the path, and with this you can *paste* it.

## Summary by CodeRabbit

* **New Features**
  * Folder picker now supports hidden-file toggling, a Home shortcut, mouse interactions, and Go To navigation.
  * Pasted paths may be quoted, absolute, home-relative, or current-folder-relative; file paths open their containing folder.
  * Search fields now accept pasted text while filtering control characters.
  * Improved terminal keyboard shortcuts, mouse interactions, diff navigation, and Git file opening.
  * Added clearer terminal lifecycle and no-workspace responses.

* **Bug Fixes**
  * Invalid pasted paths show an error without changing the current folder.
  * Copy selections now avoid unwanted margins and transcript gutters.

### DIFF
--- a/src/app/input.rs
+++ b/src/app/input.rs
@@ -680,8 +680,19 @@ impl App {
     /// all single-line fields.
     fn paste_into_modal(&mut self, s: &str) -> bool {
         use ratatui::crossterm::event::{KeyCode, KeyEvent, KeyModifiers};
+        // The folder picker has no text field to fill: a pasted path *is* the
+        // navigation. Pasting a path is how you reach a deep folder, so this is
+        // the one modal where the text means "go here" rather than "type here".
+        if self.picker.as_ref().is_some_and(|p| p.creating.is_none()) {
+            self.picker_goto(s);
+            return true;
+        }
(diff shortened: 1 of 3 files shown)


RizRiyz:
@Alex-Blanes try to rebase and fix the merge conflict issue so I can approve this.

PR #64, opened 2026-08-20
fix(agent): Claude sessions were never found on Windows

Alex-Blanes:
## Summary

- Restore Claude session discovery for Windows drive paths.
- Match the directory encoding used by Claude for separators, spaces, underscores, punctuation, and non-ASCII characters.
- Prevent a drive-relative encoded component from escaping the configured Claude session store.
- Rebase the contribution onto the current post-#63 Luvus implementation.

## Implementation

- Encode every character outside ASCII A-Z, a-z, and 0-9 as a hyphen.
- Keep the shared encoder in src/agent.rs for automatic resume, ranked session discovery, and Mission Control usage lookup.
- Route all six current duplicate-pane, hook, fork, ambiguity, and cross-tab fixtures through the production encoder.
- Preserve the newer bounded usage reader in src/agent/usage.rs instead of restoring its obsolete predecessor.
- Leave live process and sidebar state detection unchanged.

## Windows regression coverage

The focused test uses clearly synthetic paths and verifies:

- C:\Users\developer\project becomes C--Users-developer-project
- generic paths with spaces and underscores match Claude storage names
- a synthetic non-ASCII workspace name is collapsed consistently
- the resulting path stays below the configured projects directory

The regression runs in the native windows-latest CI job alongside the existing protocol and ConPTY checks.

## Local validation

- cargo fmt --all --check
- cargo clippy --all-targets -- -D warnings
- cargo test --locked
- 874 unit tests and 12 integration tests passed before the fixture-only naming cleanup

No CLI, socket, UHP, snapshot-format, or behavior changes for other agents.

## Summary by CodeRabbit

- **Tests**
  - Expanded Claude session path coverage for Windows drives, spaces, underscores, accented characters, and store containment.
  - Updated session-related tests to consistently use production path handling, including cross-tab session pairing.
  - Added the Claude project-directory encoding test to the Windows CI suite.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,6 +92,7 @@ jobs:
           cargo test --locked terminal::backend::tests
           cargo test --locked altgr
           cargo test --locked diff::git::tests::tree_tint_uses_nested_visible_spelling_and_new_rename_path
+          cargo test --locked agent::tests::claude_project_dir_encodes_every_non_alphanumeric
       - name: Validate published protocol fixtures
         shell: pwsh
         run: |
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -437,12 +437,16 @@ fn fx_base() -> PathBuf {
 // Conversations live at `<base>/projects/<encoded-cwd>/<session-uuid>.jsonl`,
 // where the cwd is encoded by replacing every `/` and `.` with `-`.
 
-fn claude_project_dir(base: &Path, cwd: &Path) -> PathBuf {
+pub(crate) fn claude_project_dir(base: &Path, cwd: &Path) -> PathBuf {
     let enc: String = cwd
         .to_string_lossy()
         .chars()
         .map(|c| {
-            if matches!(c, '/' | '\\' | '.') {
+            // `:` matters on Windows: Claude itself encodes `C:\Users\me` as
+            // `C--Users-me`, and leaving the colon in also makes `join` treat the
+            // component as a *drive-relative* path, so the lookup silently
+            // escaped `base` and landed next to the process working directory.
+            if matches!(c, '/' | '\\' | '.' | ':') {
                 '-'
             } else {
                 c
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -435,23 +435,23 @@ fn fx_base() -> PathBuf {
 
 // ── Claude Code ─────────────────────────────────────────────────────────────
 // Conversations live at `<base>/projects/<encoded-cwd>/<session-uuid>.jsonl`,
-// where the cwd is encoded by replacing every `/` and `.` with `-`.
+// where the cwd is encoded by replacing every character that is not
+// ASCII-alphanumeric with `-`.
 
 pub(crate) fn claude_project_dir(base: &Path, cwd: &Path) -> PathBuf {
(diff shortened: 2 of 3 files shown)


RizRiyz:
@Alex-Blanes I haven’t had a chance to test this on Windows. Would you be able to share a short screen recording of the claude detection working after this change?

appreciate your help.

AlexBlanesAV:
Two runs, same machine, same scratch project — the only difference is the binary. The label in the top-right corner of each video says which is which.

 **v0.11.0**: after server stop and reopening, the pane comes back as a bare shell and stays one, while the sidebar still shows resume · claude · demo-app. I wait, then start claude by hand so the question goes to an agent and not to a prompt — and it is a brand-new conversation. The thread is gone.

**With this PR**: nothing has to be typed. The pane comes back inside the same conversation and Claude answers PISTACHIO.

Both runs are driven by one script through luvus's own CLI rather than by hand, so they do the same thing and nothing depends on window focus. The script only types claude when the restored pane came back without an agent, which is why the second video has no typing after the restart. Written with Claude's help, as the commits note.
https://github.com/user-attachments/assets/b0ec1a42-7ab3-4070-a7ec-5988958b774e


https://github.com/user-attachments/assets/235a089a-753e-47ff-9cb7-f8d94d27bbee

RizRiyz:
@AlexBlanesAV resume doesnt mean the status of the agents, you need to click resume or agent on the list to resume the session. 

How did you close the luvus, are you using ctrl+space q or force closing it? the video showed that you stopped the server so thats why the session wont resume on claude.

Alex-Blanes:
You're right that the AGENTS sidebar list is a manual action — and that path isn't affected by this PR, because `claude_recent` walks `projects/` and reads the real `cwd` out of each transcript instead of encoding it. That's why the entry still showed up in the v0.11.0 video while the pane came back empty.

But automatic resume on restart is separate, and it's already luvus's own design, not something this PR adds: `App::from_snapshot` starts the restored pane's shell *on* the resume command (`src/app/mod.rs:1823-1838`, "the pane opens straight into the resuming agent — nothing visibly typed"). The session it resumes is picked at save time by `persist::resolve_pane_sessions` → `agent::latest_session` → `claude_latest` → **`claude_project_dir`**. That last call is the one this PR fixes. On Windows it builds `C:-Users-…` instead of Claude's `C--Users-…`, so it resolves to nothing and the pane restores as a plain shell — which is exactly what the first video shows.

On the shutdown: I used `luvus server stop`, which takes the `should_quit` branch and falls through to the final `persist::save(&app)` (`src/ipc/server.rs:417`), so the snapshot is written. Both videos close the same way; the only difference is the binary.

Two more things the same encoder breaks on Windows, independent of resume: `session_usage`/`session_mtime` also go through `claude_project_dir`, so the dashboard's token/context/cost column is permanently "—"; and because `C:-…` is a drive-relative path, `base.join(...)` escapes `base` and drops stray `-Users-…` folders next to the process cwd — that's how I found this in the first place (the tests started reading each other's fixtures).

Happy to record a third video with `ctrl+space q` instead if that's the close path you'd rather see.

RizRiyz:
@Alex-Blanes happy to approve once you rebase and update this PR

PR #65, opened 2026-08-20
test(windows): make the suite runnable on Windows

Alex-Blanes:
Five tests fail on `main` on any Windows machine. The `windows build` CI job only builds, so none of this shows up there. None of these are luvus bugs — the two genuine ones are #63 and #64.

- **`settings_shell_choice_cycles_and_persists`** opened tab `2`, which has been **Theme** since General was added to `SettingsTab::ALL`, and clicked control row `5`, which has been the title-path toggle since two rows were inserted above `Shell`. So the test asserted a Windows-only picker while touching neither of them. The tab key is corrected and the row index is derived from `layout_rows()` so it cannot rot again.
- **`installing_replaces_only_the_legacy_managed_hook`** interpolated a temp path into a JSON string literal. A Windows path's backslashes are invalid JSON escapes, so `settings.json` never parsed and every assertion read `null`. Built through `serde_json::json!` instead.
- **`integrate_branch_merges_clean_then_flags_a_conflict`** compares file bytes against a checkout made under Git for Windows' shipped `core.autocrlf=true`. The throwaway repo now sets `core.autocrlf=false`.
- **`opening_with_editor_spawns_a_pty_pane_in_a_new_tab`** spawned `cat`. It now uses `cmd /c type` on Windows — still a real program in a real PTY on both platforms.
- **`pane_cwd_follows_cd_without_moving_its_workspace`** and **`clicking_a_pane_title_shows_the_real_command`** assert behaviour that only exists on unix: `platform::process_cwd` and `platform::process_tree` have no Windows implementation and return `None` / an empty vec by design. Gated `#[cfg(unix)]` rather than failing where the feature deliberately isn't there.

Worth considering separately: making the Windows CI job run `cargo test` too, now that it can pass.

## Summary by CodeRabbit

* **Tests**
  * Improved cross-platform test reliability on Windows and Unix systems.
  * Updated editor and shell tests to use platform-appropriate commands and more resilient UI selection.
  * Standardized temporary Git repositories to ensure consistent line endings.
  * Improved configuration test handling for platform-specific file paths.

### DIFF
--- a/src/app/files.rs
+++ b/src/app/files.rs
@@ -718,9 +718,10 @@ mod tests {
         let mut app = App::new(120, 40, tx).unwrap();
         let tabs_before = app.workspaces[app.active_ws].tabs.len();
 
-        // `cat` stands in for an editor: a real program launched with the file as
-        // a literal argv element (present on every unix CI runner).
-        app.open_file_in_editor(file.clone(), "cat");
+        // A real program launched with the file as a literal argv element, so
+        // the pane is a true PTY: `cat` on unix, `cmd /c type` on Windows.
+        let editor = if cfg!(windows) { "cmd /c type" } else { "cat" };
+        app.open_file_in_editor(file.clone(), editor);
 
         assert_eq!(
             app.workspaces[app.active_ws].tabs.len(),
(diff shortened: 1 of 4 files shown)


RizRiyz:
@coderabbitai review

RizRiyz:
@Alex-Blanes @AlexBlanesAV does this test still need reviews?

PR #70, opened 2026-08-20
feat(git): add in-tab file diff viewer for Status section

ngurajeka:
View unified diff for staged and changed files directly in the git tab. Press Enter or d on a Status file row, or click it, to open the diff. Esc/q closes, j/k and arrows scroll, PgUp/PgDn page, g/G top/bottom.

Also adds PgUp/PgDn to commit detail and issue detail views for consistency.

image
image

image

## Summary by CodeRabbit

- **New Features**
  - Open explicitly selected changed files from Git status in the native diff viewer.
  - Added staged and unstaged file selection with keyboard and mouse controls.
  - Added diff menus, scrolling, source/note gestures, and folder-picker actions.
  - Git tabs can use the focused pane’s current directory when needed.

- **Bug Fixes**
  - Improved selection visibility and validation after status refreshes.
  - Improved mouse hit-testing and copy selection behavior.
  - PageUp/PageDown navigation now uses consistent page sizing.
  - Added notifications when no file is selected.


RizRiyz:
@ngurajeka Thanks, this is a useful addition and the CI is green. I found one issue that should be fixed before merge.

If a user opens diff A, closes it before the background `git diff` finishes, then opens diff B, A’s delayed response can overwrite B. `GitPayload::FileDiff` does not include a request or file identity, and the result is applied whenever any file diff is open.

Please attach a request token or `(path, staged)` identity to the async payload and only apply it when it matches the currently open diff.

Also, please preserve the Status scroll position when closing a diff. The keyboard behavior currently picks the center visible file or falls back to the first file, so it is not a true row selection yet.

RizRiyz:
@ngurajeka Thanks for the Git Status diff viewer. We already have the native DIFF review surface in `main`, so please do not add a separate `git diff` renderer or data model.

Instead, update the Status-row actions to open the existing native DIFF view for the selected file and change layer:

- Preserve whether the Status entry is staged or worktree
- Reuse the native DIFF preview, pane, or tab flow
- Keep the native DIFF limits, sanitization, layout controls, notes, and agent feedback
- Do not introduce a new `FileDiff`, `GitPayload::FileDiff`, or separate async diff loader
- Add regression tests that verify the selected Status file opens the correct native DIFF layer

This keeps one review experience and preserves Luvus memory and safety guarantees.

PR #73, opened 2026-08-21
fix(git): fall back to focused pane's cwd when workspace root isn't a repo

### LINKED ISSUE #67: `Ctrl+Space` then `g` does nothing
### Description

Per the docs (https://luvus.dev/docs/guides/git/), pressing `Ctrl+Space` then
`g` should open the Git tab for the current workspace. On my setup, nothing
happens at all: no visual change, no error message, and the session just
returns to IDLE.

### Expected behavior

The Git tab (Commits/Flow/Branches/PRs/Issues/Status dashboard) should open.

### Actual behavior

No visible reaction on screen after `Ctrl+Space` → `g`. No error output.
Immediately falls back to IDLE.

### Context

- Workspace is correctly inside a project containing a valid `.git`
  (`git status` works fine in the same directory outside of luvus)
- Every other `Ctrl+Space <key>` shortcut works fine

### Environment

- OS: WSL2 (Ubuntu) on Windows
- Shell: zsh
- luvus version: `0.11.0`

`luvus doctor` output:

luvus 0.11.0
✓ core the multiplexer (panes · tabs · agents) needs no external tools
✓ git git version 2.34.1 git tab · worktrees
✓ gh gh version 2.4.0+dfsg1 (20 GitHub PRs & issues
✓ ssh OpenSSH_8.9p1 Ubuntu-3ubun luvus --remote
✓ curl curl 7.81.0 (x86_64-pc-lin luvus module search
✗ keys Shift+Enter can't be distinguished from Enter here
↳ use Option+Enter instead, or switch to a terminal with
the keyboard protocol (Ghostty · Kitty · WezTerm · iTerm2)
All set — you're good to go. ✓

### Steps to reproduce

1. Open a workspace pointing to a project with a `.git` repo
2. Press `Ctrl+Space`, then `g`
3. Observe that no Git tab opens, no error is shown, and the session returns to IDLE

### Notes

`doctor` reports git/gh/ssh/curl all green, only the `keys` check fails
(Shift+Enter ambiguity, unrelated to this issue as far as I can tell — my
terminal doesn't support the Kitty keyboard protocol). Not sure if `Ctrl+Space g`
relies on a keybinding path affected by the same terminal limitation.

DylanDelobel:
`Ctrl+Space` `g` now opens the git tab even when the focused pane has `cd`'d into a repo the workspace root itself isn't in.

## What

- `open_git_tab` falls back to the focused pane's live cwd when the workspace root isn't a repo, instead of silently doing nothing.
- New `repo_check()` in `src/git/local.rs` keeps the failure reason instead of collapsing "not a repo" and "`git` failed to run" into the same `false`. A broken `git` (missing binary, bad `PATH`) now shows a toast instead of another silent no-op.
- `is_repo` is untouched; other callers keep using it.

## Why

Fixes #67.

`Workspace.cwd` is fixed at creation and never follows a pane's shell `cd` (only the branch indicator updates on checkout). My workspace was the default one in `$HOME`, one directory above the actual repo. I `cd`'d into the project inside a pane and expected `Ctrl+Space` `g` to pick that up, since the pane on screen was clearly inside a repo.

Instead, `open_git_tab` checked `$HOME`, correctly found no repo, and returned with no feedback at all, no toast, no tab, nothing. No keybinding bug, no terminal/protocol issue: a static workspace root versus a pane that had wandered into a repo underneath it.

Discussed in the issue thread: https://github.com/RizRiyz/luvus/issues/67#issuecomment-5358365138

## Verification

- [x] `cargo test --locked` (one pre-existing, unrelated failure: `cli::tests::docs_cli_reference_matches_help`, reproduces on `main` too)
- [x] Manual testing: reproduced the exact scenario from the issue (default `$HOME` workspace, `cd`'d into a repo inside a pane, `Ctrl+Space` `g`), confirmed broken on `main` and fixed on this branch.

## Notes

Added two regression tests: `repo_check` against a repo and a non-repo tempdir, and `open_git_tab` falling back to a focused pane's cwd when the workspace root isn't a repo.

## Summary by CodeRabbit

* **New Features**
  * Git actions can open the Git tab from the focused pane’s repository when the workspace root is not a repository.
  * Git repository detection now distinguishes valid repositories, non-repository folders, and Git failures.

* **Bug Fixes**
  * Git detection errors, including malformed configuration, are preserved and shown in visible toasts with failure details.
  * Git actions no longer incorrectly fall back to a pane repository when the workspace root check fails.

### DIFF
--- a/src/app/git.rs
+++ b/src/app/git.rs
@@ -1514,9 +1514,12 @@ mod tests {
         app.panes.get_mut(&focus).unwrap().cwd = repo.clone();
 
         app.open_git_tab(0);
-        assert!(
-            app.active_is_git(),
-            "git tab opens using the focused pane's cwd, not the stale workspace root"
+        let view = app
+            .active_git()
+            .expect("git tab opens using the focused pane's cwd, not the stale workspace root");
+        assert_eq!(
+            view.repo_root, repo,
+            "the git tab targets the focused pane's repo, not the workspace root"
         );
     }
 
--- a/src/app/git.rs
+++ b/src/app/git.rs
@@ -38,9 +38,16 @@ impl App {
             let pane_cwd = self.focused_cwd();
             if pane_cwd != ws_root {
                 let pane_check = local::repo_check(&pane_cwd);
-                if matches!(pane_check, local::RepoCheck::Repo) {
-                    root = pane_cwd;
-                    check = pane_check;
+                match pane_check {
+                    local::RepoCheck::Repo => {
+                        root = pane_cwd;
+                        check = pane_check;
+                    }
+                    // The pane cwd itself failed to check (bad permissions, a
+                    // broken `git`, ...): that's worth surfacing even if the
+                    // workspace root was just an ordinary non-repo folder.
+                    local::RepoCheck::Error(_) => check = pane_check,
+                    local::RepoCheck::NotRepo => {}
                 }
             }
         }
--- a/src/app/git.rs
+++ b/src/app/git.rs
@@ -34,7 +34,10 @@ impl App {
         let ws_root = self.workspaces[wsi].cwd.clone();
         let mut root = ws_root.clone();
         let mut check = local::repo_check(&root);
-        if !matches!(check, local::RepoCheck::Repo) {
+        // Only fall back on an ordinary "not a repo" root. A workspace-root
+        // `Error` (broken `git`, bad `PATH`, ...) must still surface as a
+        // toast, even when the focused pane happens to be a valid repo.
+        if matches!(check, local::RepoCheck::NotRepo) {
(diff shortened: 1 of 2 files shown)


RizRiyz:
@dylandelobel One small request before merging: could you strengthen `git_tab_falls_back_to_focused_pane_cwd` by asserting that the opened Git tab’s `repo_root` is exactly the focused pane’s repository path? The current test only confirms that a Git tab was opened, so it could miss a regression where the correct tab type opens with the wrong repository.

Everything else looks good, and the existing workspace-root behavior remains intact.

DylanDelobel:
`git_tab_falls_back_to_focused_pane_cwd` now asserts `GitView.repo_root` equals the focused pane's repo path👌

Issue #72, opened 2026-08-20
Reorder tabs


escobar5:
**What problem would this solve?**
Move tabs right and left

**What would you like to happen?**
Would be nice to have an option to reorder tabs, I don't see it anywhere.

**Alternatives**
I think this is core functionality

RizRiyz:
@escobar5 This feature is still in beta. We plan to improve it by adding TUI support, including a right-click option to move tabs, similar to how pane reordering works today.

Here is how to test reorder/move tab in CLI, it currenly only work in active tabs.

```
$ luvus tab move 4 8
{
  "id": "1",
  "result": {
    "active": "2",
    "from": "4",
    "to": "8",
    "type": "tab_move"
  }
}
```

PR #75, opened 2026-08-21
feat(tabs): add precise reorder and swap controls

RizRiyz:
## Summary

Adds precise tab organization across the TUI, CLI, and socket API. Tabs can now be renamed, moved one step, moved to an exact position, or swapped with another tab while preserving the active tab and every running pane process.

Refs #72

### TUI

- Right-clicking a tab now opens a context menu instead of immediately opening rename.
- **Rename** opens the existing tab-name editor for the exact tab that was clicked.
- **Move Left** and **Move Right** reorder the clicked tab by one position.
- **Swap With** lists every other tab in the workspace, including tabs hidden by tab-bar overflow, and exchanges the two selected positions.
- Tab-scoped module actions remain composable in the same context menu.
- Context-menu targets are tracked by tab identity, so an API reorder while the menu is open cannot make an action affect the wrong tab.

### CLI

```sh
luvus tab focus 4
luvus tab rename review
luvus tab rename review --tab 2
luvus tab move left
luvus tab move right --tab 4
luvus tab move 3 1
luvus tab swap 1 4
```

- `tab focus <n>` selects an exact 1-based tab.
- `tab rename <name> [--tab N]` renames the active or selected tab; an empty name clears the custom label.
- `tab move left|right [--tab N]` moves the active or selected tab one position.
- `tab move <from> <to>` moves a tab to an exact final position.
- `tab swap <first> <second>` exchanges two tab positions without focusing a different tab.

### Socket API

| Method | Parameters | Result |
|---|---|---|
| `tab.focus` | `{"tab": n}` | `{type:"ok"}` |
| `tab.rename` | `{"name":"review","tab":n?}` | `{type:"ok"}` |
| `tab.move` | `{"tab":from,"to":to}` | `{type:"tab_move",from,to,active}` |
| `tab.move` | `{"direction":"left"|"right","tab":n?}` | `{type:"tab_move",from,to,active}` |
| `tab.swap` | `{"tab":first,"with":second}` | `{type:"tab_swap",tab,with,active}` |

## Behavior and safety

- All public tab positions are strict, positive, 1-based values.
- Missing, zero, malformed, out-of-range, same-position, and edge moves return `invalid_request` without partially changing state.
- Non-null `direction` cannot be combined with `to`; `direction: null` is treated as omitted so positional move requests remain valid.
- Direction parsing accepts only `left` and `right`.
- Reordering and swapping preserve the identity of the active tab even when its numeric position changes.
- Pane layouts, PTYs, running commands, and agent sessions stay attached to their original tab objects.
- Dashboard tabs keep their fixed labels and cannot be renamed.
- Tab names are trimmed, can be cleared, and are limited to 40 characters.

## Localization and documentation

- Adds localized tab-menu labels and validation toasts for all eight supported languages.
- Updates the layout guide and CLI/API references with move, swap, focus, and rename examples.

## Verification

- `cargo test --all-targets` — 593 unit tests, 7 CLI tests, and 2 named-session tests passed.
- `cargo clippy --all-targets -- -D warnings` — passed.
- `cargo fmt --all -- --check` and `git diff --check` — passed.
- `npm run build` in `website/` — passed.

## Scope

- No pane processes are restarted or recreated by tab organization operations.
- No unrelated files are included in the branch diff.


RizRiyz:
https://github.com/user-attachments/assets/2fd72eb0-e1f3-4fed-8aa3-397d70944718

Issue #106, opened 2026-08-24
Make the VT engine boundary pluggable: kind-open factory and an out-of-process engine transport


franchb:
## Summary

#49 already commits to keeping `VtEngine` as the boundary and to adding "an
engine kind and one factory responsible for validation and construction". This
issue proposes landing that factory **engine-kind-open**, plus one
transport-based engine implementation, so alternative engines can be built and
measured out-of-tree without Luvus vendoring any of them.

The goal is not a second engine in Luvus. It is that the seam #49 already
promises becomes usable from outside the crate.

## Problem

Today an alternative engine cannot be prototyped without forking Luvus:

- `luvus` is a binary-only crate — there is no `src/lib.rs`, so
  `crate::terminal::vt::VtEngine` (`src/terminal/vt/mod.rs`) is unreachable from
  any external crate.
- `/plugins` is excluded from the published crate, and modules
  (MODULE-GUIDE.md) are action, dock and pane extensions — they cannot host a
  terminal engine.
- Every construction site names the concrete type directly
  (`src/terminal/pty.rs:407`, `src/terminal/pty.rs:480`, `src/main.rs:1878`,
  `src/ipc/api.rs:2176`).

So the cost of evaluating any engine — including the one in #49 — is a fork plus
a full build integration, before a single benchmark number exists. That is
backwards: the gates in #49 (p95 within 5%, memory within 10%, reproducible
builds) are exactly the things you want measurable *before* committing to an
integration.

## Why a process boundary rather than FFI

#49's plan links a native engine into the Luvus process. That places every cost
on Luvus:

- `unsafe` FFI with lifetime, callback and teardown handling;
- a native toolchain in `build.rs` that must stay optional on every platform;
- per-platform reproducible builds, including Nix, for a foreign build system;
- licence review of a linked native library;
- a threading model Luvus does not control, inside `Arc<Mutex<dyn VtEngine>>`.

An out-of-process engine removes all five. Luvus gains no `unsafe`, no native
toolchain, no vendored source, and no linking relationship. The engine is just a
program Luvus talks to, in any language, exactly like modules.

## Design sketch

- `VtEngineKind` with the default `Alacritty` (in-process, unchanged) and one
  `External` kind carrying a command plus arguments.
- One factory owning validation and construction, as #49 already specifies.
  Missing configuration selects Alacritty; an invalid value is an actionable
  error; an engine that fails to start is reported, never silently downgraded.
[...]

Issue #49, opened 2026-08-16
Offer an experimental Ghostty VT engine alongside Alacritty


RizRiyz:
## Summary

Add `libghostty-vt` as an experimental terminal-engine option while retaining Alacritty as Bohay’s stable default. Each pane must instantiate exactly one backend, chosen by a server-wide setting and applied after restart.

## User experience

- Add a **VT engine** setting under Terminal with **Alacritty (stable)** and **Ghostty (experimental)** choices.
- Explain that the change takes effect after a Bohay server restart.
- Keep existing live panes on their current engine; never migrate a live grid, scrollback, parser state, or alternate screen between engines.
- Show active, configured, and unavailable engines in diagnostics.
- Missing configuration continues to select Alacritty. Invalid values produce an actionable error.
- If Ghostty is unavailable in a build or fails to initialize, report that clearly. Do not silently fall back to Alacritty.

## Architecture

- Preserve the existing `VtEngine` abstraction as the boundary used by panes, UI, and IPC.
- Add an engine kind and one factory responsible for validation and construction.
- Implement the Ghostty adapter behind an optional build feature; a normal source `cargo build` must remain Alacritty-only and not require Zig.
- Official artifacts can include both engines after reproducible, cross-platform build support is proven.
- Allocate only the selected engine for a pane; no hidden standby grid or doubled per-pane memory.

## Compatibility and safety requirements

The Ghostty adapter needs behavior parity for terminal input, resize/reflow, cells and styling, visible text, scrollback, alternate screen, terminal modes, mouse reporting, bracketed paste, titles, terminal-generated responses, and Bohay’s agent UI detection. Keep FFI and unsafe code isolated, with explicit lifetime, callback, error, and teardown handling.

Create a shared backend conformance suite covering ASCII and Unicode, wrapping, resize, primary and alternate screens, scrollback, SGR styles, OSC titles, terminal responses, input modes, malformed escape sequences, and supported agent fixtures.

## Performance and packaging gates

Benchmark optimized builds with identical pane sizes, scrollback limits, and workloads. Record parsing/state work, grid-to-render-cell traversal, and Bohay rendering/socket delivery separately. Measure p50 and p95 for plain output, color-heavy agent output, Unicode, full-screen redraws, resize/reflow, scrollback, nine panes, and repeated pane churn.

Before exposing the selector in release builds, verify:

- no measurable idle CPU regression;
- no common-path per-frame allocation regression;
[...]

PR #109, opened 2026-08-24
refactor(terminal): construct VT engines through one factory

franchb:
## What

Route VT engine construction through one factory instead of naming the concrete
engine at each call site.

- `VtEngineKind` — one variant, `Alacritty`, and it is the `Default`.
- `create_engine(kind, cols, rows, resp_tx, history_budget_bytes)` — returns
  `Arc<Mutex<dyn VtEngine>>`, which is what both callers already built by hand.

Both production sites in `src/terminal/pty.rs` (pane spawn and deferred pane
construction) now call the factory. The two remaining direct constructions are
in `#[cfg(test)]` code that deliberately drives a concrete engine, so they are
left alone.

## Why

`VtEngine` is already the boundary the rest of the app talks to, but the choice
of engine was spelled out wherever a pane was built. That makes the decision
implicit and duplicated: two identical seven-line constructions that must be
kept in sync, and no single place to put validation, diagnostics or a future
alternative.

#49's architecture section already calls for exactly this — "add an engine kind
and one factory responsible for validation and construction" — so this is that
step, extracted on its own and with no engine attached to it. Doing it
separately keeps the refactor reviewable and keeps the interesting work in #49
free of unrelated churn.

Related: #106, which proposes what the seam could carry once it exists. This PR
takes no position on that; it is a straight refactor either way.

## No behavior change

One kind, every caller passes `VtEngineKind::default()`, and the engine built is
the same `AlacrittyEngine` with the same arguments. Nothing is configurable and
no user-visible surface moves.

## Tests

`terminal::vt::tests::factory_builds_a_working_default_engine` — builds through
the factory, feeds bytes, and asserts the grid and cursor, so the factory is
covered rather than merely compiled.

Full local run against the CI gates:

- `cargo fmt --package luvus --check` — clean
- `cargo test --locked` — 834 passed, 0 failed

## Summary by CodeRabbit

* **New Features**
  * Added configurable terminal engine selection while preserving the default Alacritty-based experience.
  * Restored terminal panes can now replay saved screen content during startup.
  * Added reliable detection of pending terminal data.

* **Bug Fixes**
  * Improved startup reliability by retrying the requested and fallback working directories.
  * Preserved pasted content containing Windows paths, quotation marks, and Unicode characters.
  * Improved terminal output parsing and cursor updates.

### DIFF
--- a/src/terminal/pty.rs
+++ b/src/terminal/pty.rs
@@ -16,8 +16,7 @@ use portable_pty::{native_pty_system, CommandBuilder, MasterPty, PtySize};
 use crate::event::AppEvent;
 use crate::ids::PaneId;
 use crate::terminal::backend::TerminalRuntime;
-use crate::terminal::vt::alacritty::AlacrittyEngine;
-use crate::terminal::vt::VtEngine;
+use crate::terminal::vt::{create_engine, VtEngine, VtEngineKind};
 
 const CHILD_REAPER_INTERVAL: Duration = Duration::from_millis(50);
 
@@ -475,12 +474,13 @@ impl Pane {
         // All bytes (user input + terminal responses) funnel through one channel
         // to a single writer thread — keeps ordering correct, needs no mutex.
         let (input_tx, input_rx) = mpsc::channel::<InputAction>();
-        let engine: Arc<Mutex<dyn VtEngine>> = Arc::new(Mutex::new(AlacrittyEngine::new(
+        let engine = create_engine(
+            VtEngineKind::default(),
             cols,
             rows,
             input_tx.clone(),
             history_budget_bytes,
-        )));
+        );
         // Replay the saved screen so a restored pane shows its prior content.
         if let Some(screen) = initial {
             if let Ok(mut e) = engine.lock() {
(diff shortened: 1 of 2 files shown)


RizRiyz:
@franchb are you plan to bring multiple vtengine into luvus with this? since this is a construction-site cleanup, not a product change, its possible to merge this PR.

franchb:
Yes — that's where I'd like it to go, and I wrote it up as #106 so the direction is visible rather than implied.

The short version: I want to run a different VT core behind this seam and measure it against Alacritty — parse throughput, memory per retained row, and a conformance diff — without Luvus vendoring it or taking on a native toolchain. The engine I'm prototyping is Shitty's, and I'd run it out-of-process rather than linked, precisely so its C++ toolchain, threading model and licensing stay off your side of the boundary.

This PR deliberately carries none of that, though: one kind, every caller passes the default, same engine with the same arguments. If #106 goes nowhere it's still the "engine kind and one factory" step #49 already asks for, and the duplicated construction is gone either way — happy for it to be judged purely as cleanup.

RizRiyz:
@franchb can you rebase and resolve the conflicts?

franchb:
Rebased onto `main` (08ac195). The conflict was with #116: `build_deferred` gained an initial-screen replay right after engine construction, so the resolution keeps that replay and simply feeds it the engine the factory returns.

Re-ran the gates after the rebase: `cargo fmt --check` clean, `cargo clippy --all-targets -- -D warnings` clean, `cargo test --locked` 847 passed.

#115 is stacked on this and has been rebased on top of it as well.

PR #115, opened 2026-08-24
test(terminal): pin cell-level VT engine behaviour

franchb:
> **Stacked on #109.** Until that merges this PR shows both commits; the one to
> review here is `test(terminal): pin cell-level VT engine behaviour`. GitHub
> will narrow the diff automatically once #109 lands.

## Problem

The existing engine tests assert rendered *text*. That cannot distinguish a
grapheme cluster held in one wide cell from the same codepoints split across two
cells — both render the same string. But the two place every following column
differently, so the difference is invisible to the tests and plainly visible in a
pane.

It is also easy to change by accident: a bump of `alacritty_terminal`, a
`unicode-width` update, or any future engine work can move a cluster's width
without touching a line of Luvus.

## Change

Seven characterisation tests over `for_each_cell`, asserting *where each cluster
lands* rather than what the row reads as. The helper dumps the visible grid as
`col:CODEPOINT+CODEPOINT` tokens and omits blanks, so a cluster that grew wider
shows up as a gap in the column numbers.

Covered: ASCII placement, wide characters, combining marks, VS16, emoji ZWJ
sequences, emoji modifier sequences, and soft wrap at the right margin.

They are written against `VtEngine`, not against `AlacrittyEngine`, so they hold
for whatever `create_engine` returns.

## What they pin

Three of these are worth knowing about, because they are decisions the engine
makes on Luvus's behalf and agent CLIs emit these characters constantly:

| input | current behaviour |
|---|---|
| `👩‍💻` `U+1F469 U+200D U+1F4BB` | joiner stays with the first emoji, second emoji gets its own wide cell — **4 columns** |
| `👍🏽` `U+1F44D U+1F3FD` | modifier takes its own wide cell — **4 columns** |
| `❤️` `U+2764 U+FE0F` | **1 column** |

UTS #51 treats an emoji ZWJ sequence and an emoji-modifier sequence as single
width-2 clusters, so a strictly conforming engine would place these differently.
I am not proposing a change — the tests pin what Luvus does today, with the
divergence noted in a comment so the choice is visible rather than accidental.

## Relationship to #49

#49 asks for a shared backend conformance suite before a second engine can be
called equivalent. This is the smallest useful piece of it, contributed with no
engine attached: it works today as regression protection for the engine already
here, and any future engine has to match it or explain why.

## Tests

- `cargo fmt --package luvus --check` — clean
- `cargo test --locked` — 841 passed, 0 failed (834 before, 7 added)

## Summary by CodeRabbit

* **New Features**
  * Added support for restoring terminal screens, including saved content and paste state.
  * Improved deferred terminal launches with fallback working directories and clearer startup status.
  * Added support for detecting pending terminal output without consuming it.

* **Bug Fixes**
  * Improved terminal rendering for wide characters, combining marks, variation selectors, and emoji sequences.
  * Fixed edge cases involving soft wrapping and character placement.
  * Improved terminal pane initialization for more consistent behavior across launch paths.

### DIFF
--- a/src/terminal/vt/mod.rs
+++ b/src/terminal/vt/mod.rs
@@ -246,3 +246,103 @@ mod tests {
         assert_eq!(engine.cursor().x, 2);
     }
 }
+
+#[cfg(test)]
+mod conformance {
+    //! Characterisation of the `VtEngine` contract at the cell level.
+    //!
+    //! These assert *where each grapheme cluster lands*, not just the text a
+    //! row renders to. A text comparison cannot see the difference between a
+    //! cluster held in one wide cell and the same codepoints split across two,
+    //! yet that difference moves every column after it on the line — so it is
+    //! exactly what a text-level test misses and a pane visibly gets wrong.
+    //!
+    //! They are written against the trait rather than any engine, so they hold
+    //! for whatever backs `create_engine`.
+
+    use super::*;
+    use std::sync::mpsc;
+
+    /// Visible grid as one `"col:CODEPOINT+CODEPOINT"` token per occupied cell,
+    /// row by row. Blank cells are omitted, so a token's column is the
+    /// assertion: a cluster that grew wider shows up as a gap.
+    fn cell_dump(input: &[u8], cols: u16, rows: u16) -> Vec<String> {
+        let (tx, _rx) = mpsc::channel();
+        let engine = create_engine(VtEngineKind::default(), cols, rows, tx, 64 * 1024);
+        let mut engine = engine.lock().expect("engine lock");
+        engine.advance(input);
+
+        let mut out: Vec<Vec<String>> = vec![Vec::new(); rows as usize];
+        engine.for_each_cell(&mut |row, col, symbol, _style| {
+            if symbol == " " {
+                return;
+            }
+            let points: Vec<String> = symbol
+                .chars()
+                .map(|ch| format!("{:X}", ch as u32))
+                .collect();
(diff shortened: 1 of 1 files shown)


franchb:
Rebased onto the updated #109, which is itself now rebased onto `main` (08ac195). No conflicts in this commit — it only adds tests.

Gates after the rebase: `cargo fmt --check` clean, `cargo clippy --all-targets -- -D warnings` clean, `cargo test --locked` 854 passed.

RizRiyz:
@franchb I have merged the #109 thanks for ur contribution.

Also happy to hear the research progress on new vte, feel free to post it in issues.

franchb:
Rebased onto `main` now that #109 has merged, so this is down to its own commit — 100 lines, tests only, no conflicts.

`cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean; `cargo test --locked` passes.

franchb:
Following your decision on #106 — I've closed that, and this PR stands entirely apart from it.

Nothing here references a second engine or the bindings. It is characterisation of the engine Luvus ships today: seven tests over `for_each_cell` asserting where each grapheme cluster lands, which a text-level assertion cannot see. What they pin is that `👩‍💻` and `👍🏽` each take four columns and `❤️` takes one — decisions `alacritty_terminal` makes on Luvus's behalf, that an upstream bump or a `unicode-width` change could move without touching a line of Luvus.

Rebased onto `main`, one commit, 100 lines, tests only. Happy to close it if you would rather not carry them.

PR #120, opened 2026-08-25
feat(tabs): new tab and split inherit focused pane's cwd

sinardyas:
Opening a new tab or split now starts in the directory you're working in, not the workspace root. A General setting keeps the old reset-to-root behavior. New tabs and splits also survive their target directory being deleted before the shell starts.

## What

- New tabs inherit the focused pane's live cwd instead of the workspace root.
- One shared resolver backs both `new_tab` and `split`. It returns the ordered candidate directories that still exist (focused pane's cwd, then workspace root, then `$HOME`), not a single path. When none survive it anchors on `$HOME`, so a deleted root is never handed back (the old code returned it via `unwrap_or(root)`).
- `split` spawns deferred, so it now passes the whole chain to the PTY worker. The worker tries each candidate and re-checks existence at fork time, so a directory deleted between resolving the cwd and forking the shell falls through instead of spawning a dead pane (`PtyExit`).
- A new General setting, "Open new pane/tab at workspace root" (`layout.new_pane_to_workspace_root`, off by default), makes reset-to-root opt-in. Translated in all 8 languages.
- The workspace root itself is unchanged. The change lives in the shared resolver, so the keybind and the `tab.new` API both get it.

## Why

`new_tab` reset you to the workspace root while `split` and `new_workspace` already inherited the focused pane's cwd. This aligns `new_tab` with them.

It also fixes two findings from review. The resolver returned the workspace root even after it failed the existence check, handing a new pane a directory already gone. And the deferred split passed only the single resolved cwd with no fallbacks, so a directory removed in the race window before the fork could make the split silently disappear. The resolver now returns the full existence-filtered chain and the split threads it to the worker, matching the session-restore path, which already retries a fallback list.

The toggle keeps the clean reset available for people who preferred it, without making it the default.

## Verification

- [ ] `cargo test --locked`
- [ ] `cargo clippy --all-targets -- -D warnings`
- [ ] `cargo fmt --all --check`
- [ ] Manual testing, if applicable

## Notes

- Live cwd is tracked on macOS/Linux by process inspection (`proc_pidinfo` / `/proc/{pid}/cwd`, not OSC 7). On Windows and elsewhere `process_cwd` returns `None`, so a new tab inherits the focused pane's spawn directory. Still not the root.
- The cwd is polled about once a second, so a `cd` immediately followed by new-tab can briefly inherit the previous directory. Accepted, and it matches how `split` already behaves.
- The in-app toggle takes effect on the next new tab. A hand-edited `config.json` needs a restart or a `server.reload_config` call (no file watcher).
- Two regression tests cover the fix: the deferred worker retries a fallback cwd, and the resolver never returns a missing directory.
- Verified locally: `cargo test` passes (849) and `cargo fmt --all --check` is clean. `cargo clippy -D warnings` flags two pre-existing `nonminimal_bool` lints in `src/ui/menu.rs`, which this PR does not touch.

## Summary by CodeRabbit

* **New Features**
  * Added a General setting to open new panes at the workspace root.
  * The setting is disabled by default, persists across sessions, and is available in multiple languages.
  * When disabled, new panes prefer the focused pane’s directory.

* **Bug Fixes**
  * New panes now fall back through available directories, including the workspace root and home directory, when the preferred location is unavailable.
  * Deferred pane creation now retries fallback locations automatically.


RizRiyz:
@sinardyas resetting to the workspace root sounds like good idea but this should be an optional for users, can you point this feature to settings > general > New pane/tab to workspace root? so user can select to enable this feature if they want the reset to workspace root. 

That idea will give better option for our users.

sinardyas:
> @sinardyas resetting to the workspace root sounds like good idea but this should be an optional for users, can you point this feature to settings > general > New pane/tab to workspace root? so user can select to enable this feature if they want the reset to workspace root.
> 
> That idea will give users better option for out users.

ah that would be the better approach. let me add it

RizRiyz:
@sinardyas Thanks, the new General setting and persistence behavior look good.

The two cwd fallback findings are still present, though:

- `spawn_cwd()` still returns the workspace root when every candidate is invalid.
- Deferred splits still pass only one cwd to the PTY worker, so a directory removed after resolution can make the split
disappear without trying workspace root or home.

Could you preserve the fallback candidates through deferred spawning, handle the no-valid-directory case cleanly, and add regression tests for both cases?

Also the docstring coverage warning is not worth addressing here.

PR #134, opened 2026-08-26
fix(mission): price Fable/Mythos 5, refresh Anthropic rates, and use their real 1M window

### LINKED ISSUE #133: Mission Control: no cost and ~5× inflated context for Claude Fable/Mythos 5; stale Opus/Haiku prices
## Summary

Two Mission Control accuracy gaps for current Anthropic models (v0.13.1, `src/mission/pricing.rs`):

1. **Cost shows `—` for Claude Fable/Mythos 5.** `model_price` substring-matches only `opus`/`sonnet`/`haiku` (plus OpenAI families), so `claude-fable-5` / `claude-mythos-5` sessions fall through to `None`. The Anthropic rows are also stale: Opus is priced at $15/$75 per MTok (Opus 4.1-era) while every current Opus (4.6 through 5) is $5/$25, and Haiku at $0.8/$4 vs. Haiku 4.5's $1/$5.
2. **Context % reads ~5× too full for always-1M models.** `model_window` assumes a 200k Claude window, and `context_frac` only infers the 1M window once `ctx_tokens > 200_000`. Claude Fable/Mythos 5 run a 1M window unconditionally (it is also the default), so any session under 200k context — i.e. most of a session's life — renders as up to 5× fuller than it is (a real 12% shows as 60%), and trips the ≥85% "compacts soon" flag far too early.

`config.mission_pricing` can patch (1) locally, but there is no override for the window, so (2) is not user-fixable.

## Repro

Open Mission Control on a live Claude Code session running `claude-fable-5`:
- COST column: `—`
- CONTEXT: compare against Claude Code's own context meter (% of 1M) — Mission Control shows ~5× the real fraction while under 200k.

## Suggested fix

Add a `fable`/`mythos` arm to both tables: `(10.0, 50.0, 1.0)` in `model_price` (current first-party rate, cache at the standard 0.1× read multiplier) and `1_000_000` in `model_window`; refresh the Opus/Haiku rows to current rates. The `>200k ⇒ 1M` inference in `context_frac` stays as-is for the opt-in-1M Opus/Sonnet families.

PR with this change (tests included) incoming shortly.

zengbo:
Mission Control now prices Claude Fable/Mythos 5 sessions and shows their real context fraction, instead of a `—` cost and a ~5× inflated context bar.

## What

- `model_price`: adds a `fable`/`mythos` arm at the current first-party rate ($10/$50 per MTok, cache read at the standard 0.1× input), and refreshes the stale Anthropic rows — Opus $15/$75 → $5/$25 (every current Opus, 4.6 through 5), Haiku $0.8/$4 → $1/$5 (Haiku 4.5). Sonnet stays $3/$15 (Sonnet 4.6). `config.mission_pricing` overrides keep working unchanged.
- `model_window`: `fable`/`mythos` → 1,000,000. These models run a 1M window unconditionally (it is also their default), so the 200k base + `>200k ⇒ 1M` inference in `context_frac` read every sub-200k session as up to 5× fuller than it was and tripped the ≥85% "compacts soon" flag far too early. The inference path is untouched for the opt-in-1M Opus/Sonnet families (the existing 978k ⇒ 98% test still passes).
- Tests: Fable is priced (not `—`), Mythos matches, and `context_frac("claude-fable-5", 100_000)` reads 10% rather than 50%.

## Why

Fixes #133. Live sessions on `claude-fable-5` showed no COST and a CONTEXT bar ~5× above Claude Code's own meter while under 200k context.

## Verification

- [x] `cargo test --locked` — 919 passed; the one failure, `app::tests::resize_yields_to_pane_title_and_zoom_but_still_grabs_the_seam`, fails identically on unmodified `main` in this environment (headless aarch64 Linux) and is unrelated
- [x] Manual testing: equivalent values applied via `config.mission_pricing` on a live v0.13.1 server show the expected cost for a `claude-fable-5` session

https://claude.ai/code/session_01TUVcFgNki8KP8VWsm9zvSK

## Summary by CodeRabbit

* **New Features**
  * Added support for bare Claude `fable`, `opus`, `sonnet`, and `haiku` aliases with current pricing.
  * Added support for `latest` suffixes on versioned Claude model IDs.
  * Applied conservative 200k-token context windows to aliases, except `fable`, which supports 1M tokens.

* **Bug Fixes**
  * Improved Claude model validation to reject malformed suffix combinations.
  * Refined pricing and context-window handling for Claude aliases and versioned models.

### DIFF
--- a/src/mission/pricing.rs
+++ b/src/mission/pricing.rs
@@ -7,41 +7,146 @@
 /// near this line are flagged "compacts soon". Matches the orch gate.
 pub const COMPACT_AT: f32 = 0.85;
 
-/// Rough USD price per **million** tokens as `(input, output, cache)` for `model`
-/// (substring match on the model id). `None` for an unknown model, so its cost is
-/// shown as "—" rather than a wrong number. Estimates only; overridable via config
-/// (MC-5). Kept deliberately conservative and easy to eyeball. Anthropic rows
-/// track the current first-party API rates (2026-08): Fable/Mythos 5 at
-/// $10/$50, Opus 4.6–5 at $5/$25, Haiku 4.5 at $1/$5, with cache at the
-/// standard 0.1× input read rate.
+enum ClaudeModel<'a> {
+    Versioned {
+        family: &'a str,
+        version: (u32, u32),
+    },
+    MythosPreview,
+}
+
+fn is_snapshot(part: &str) -> bool {
+    part.len() == 8 && part.bytes().all(|byte| byte.is_ascii_digit())
+}
+
+fn valid_model_suffix(parts: &[&str]) -> bool {
+    let parts = if parts.first().is_some_and(|part| is_snapshot(part)) {
+        &parts[1..]
+    } else {
+        parts
+    };
+    match parts {
+        [] => true,
+        [provider] => provider.strip_prefix('v').is_some_and(|version| {
+            !version.is_empty() && version.bytes().all(|b| b.is_ascii_digit())
+        }),
+        [provider, revision] => {
+            provider.strip_prefix('v').is_some_and(|version| {
+                !version.is_empty() && version.bytes().all(|b| b.is_ascii_digit())
+            }) && revision.bytes().all(|byte| byte.is_ascii_digit())
+        }
+        _ => false,
+    }
+}
+
+fn parse_generation(part: &str) -> Option<u32> {
+    if is_snapshot(part) {
+        None
+    } else {
+        part.parse().ok()
+    }
+}
+
+/// Parse a Claude family and `(major, minor)` version from first-party or cloud
+/// model ids. Before Claude 4.6, most ids put the family before the version and
+/// append a date, while Claude 3.5 Haiku used `claude-3-5-haiku-<date>`.
+fn claude_model(model: &str) -> Option<ClaudeModel<'_>> {
+    let parts: Vec<_> = model
+        .split(|c: char| !c.is_ascii_alphanumeric())
+        .filter(|part| !part.is_empty())
+        .collect();
+    let claude_index = parts.iter().rposition(|part| *part == "claude")?;
+    if claude_index != 0 && parts.get(claude_index - 1) != Some(&"anthropic") {
+        return None;
+    }
+    let model = &parts[claude_index + 1..];
+    if model.starts_with(&["mythos", "preview"]) {
+        return valid_model_suffix(&model[2..]).then_some(ClaudeModel::MythosPreview);
+    }
+
+    let (family, version, suffix) =
+        if let Some(major) = model.first().and_then(|part| parse_generation(part)) {
+            // Claude 3.5 Haiku and other Claude 3 ids put the version first.
+            if let Some(minor) = model.get(1).and_then(|part| parse_generation(part)) {
+                (*model.get(2)?, (major, minor), &model[3..])
+            } else {
+                let family = *model.get(1)?;
+                if is_snapshot(family) {
+                    return None;
+                }
+                (family, (major, 0), &model[2..])
+            }
+        } else {
+            let family = *model.first()?;
+            let major = parse_generation(model.get(1)?)?;
+            if let Some(minor) = model.get(2).and_then(|part| parse_generation(part)) {
+                (family, (major, minor), &model[3..])
+            } else {
+                (family, (major, 0), &model[2..])
+            }
+        };
+    valid_model_suffix(suffix).then_some(ClaudeModel::Versioned { family, version })
+}
+
+/// Rough USD price per **million** tokens as `(input, output, cache)` for `model`.
+/// `None` for an unknown model, so its cost is shown as "—" rather than a wrong
+/// number. Estimates only; overridable via config (MC-5). Kept deliberately
+/// conservative and easy to eyeball. Anthropic rows track the current first-party
+/// API rates (2026-08): Fable/Mythos 5 at $10/$50, Opus 4.5–5 at $5/$25,
+/// Haiku 4.5 at $1/$5, with cache at the standard 0.1× input read rate.
 pub fn model_price(model: &str) -> Option<(f64, f64, f64)> {
     let m = model.to_lowercase();
-    let p = if m.contains("fable") || m.contains("mythos") {
-        (10.0, 50.0, 1.0)
-    } else if m.contains("opus") {
-        (5.0, 25.0, 0.5)
-    } else if m.contains("sonnet") {
-        (3.0, 15.0, 0.3)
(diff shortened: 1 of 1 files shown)


RizRiyz:
@zengbo Thanks for working on this. The Fable and Mythos correction is valid and the PR is nicely focused. It fixes their missing cost and incorrect context percentage.

Before merging, could you please make the model matching version-aware instead of using broad substring checks?

- `contains("opus")` currently changes legacy Opus models such as Opus 4.1 from their correct $15/$75 rate to $5/$25
- `contains("haiku")` changes Haiku 3.5 from $0.80/$4 to $1/$5
- Please distinguish Mythos 5 from Mythos Preview where needed
- Please add fixed 1M context windows for the other current models that use them, including Opus 4.6 and later and Sonnet 4.6 and later
- Please add regression tests covering both current and legacy model IDs
- Fix suggestion from coderabbit above

The existing `mission_pricing` override behavior should remain unchanged.

Official references:

- [Anthropic model pricing](https://platform.claude.com/docs/en/about-claude/pricing)
- [Anthropic context windows](https://platform.claude.com/docs/en/build-with-claude/context-windows)

Once those cases are covered, this should be a solid Mission Control fix. Thank you.

zengbo:
Thanks for the review! Addressed in 8a9287e:

- **Version-aware matching**: added a small `claude_model` parser that extracts the family and `(major, minor)` version from first-party and cloud-style ids (`claude-opus-4-1-20250805`, `anthropic.claude-...-v1:0`, and the older `claude-3-5-haiku-<date>` ordering), instead of substring checks. Unparseable ids fall through to `None` so cost renders as "—" rather than a wrong number.
- **Legacy rates preserved**: Opus ≤4.1 stays $15/$75/$1.50, Haiku 3.5 stays $0.80/$4/$0.08; $5/$25 applies to Opus ≥4.5 and $1/$5 to Haiku ≥4.5 only. Sonnet 5 is priced at its $2/$10 rate (now standard per the pricing page note), Sonnet ≤4.6 keeps $3/$15.
- **Mythos Preview vs Mythos 5**: distinguished — Preview is $25/$125 per [anthropic.com/glasswing](https://anthropic.com/glasswing), Mythos 5 shares Fable 5's $10/$50.
- **1M windows**: fixed 1M now covers Opus ≥4.6, Sonnet ≥4.6, Fable/Mythos 5, and Mythos Preview, matching the [context-windows doc](https://platform.claude.com/docs/en/build-with-claude/context-windows) ("Claude 4.6 and later models and Claude Mythos Preview"). Earlier models keep the existing 200k + inference behavior.
- **Regression tests**: added for current and legacy ids on both price and window (14 new assertions), including the CodeRabbit case — `claude-mythos-preview` embedded in a longer id no longer matches. `cargo test` mission::pricing: 4 passed, 0 failed.
- **`mission_pricing` override**: untouched.

All rates cross-checked against the [official pricing page](https://platform.claude.com/docs/en/about-claude/pricing) today. Happy to adjust further if anything else stands out.

RizRiyz:
Wonderful @zengbo!! Thanks for addressing the version-aware pricing issue. Could you also fix the remaining alias regression before we merge?

The stricter parser correctly avoids broad substring matching, but it now returns no price for valid aliases that were previously handled, including `claude-3-5-sonnet-latest` and Claude Code family aliases such as `opus`, `sonnet`, and `haiku`.

Please keep the fix narrow:

- accept `latest` only after a recognized versioned Claude model
- handle supported family aliases explicitly instead of using broad substring matching
- preserve the existing version-specific and preview pricing
- keep unknown or malformed model names unpriced
- add focused regression tests for `claude-3-5-sonnet-latest`, `opus`, `sonnet`, `haiku`, `claude-opus`, and invalid suffixes

Because Claude Code aliases can resolve differently by provider and may change over time, please avoid assigning current pricing to arbitrary strings that merely contain a family name. Keep context-window handling conservative when the resolved version is unknown.

After that focused fix and passing CI, PR #134 should be ready to merge.

zengbo:
Thanks — the alias regression is fixed in c79133d (only `src/mission/pricing.rs` changed):

- **`latest` suffix**: accepted only after a recognized versioned Claude id — `claude-3-5-sonnet-latest` now prices and windows exactly as `claude-3-5-sonnet`. It composes with the existing suffix validation, so `claude-mythos-preview-latest` and `claude-3-5-sonnet-latest-v1` stay unpriced.
- **Family aliases**: `opus`, `sonnet`, `haiku` are handled as an explicit exact-match `FamilyAlias` arm (no substring matching anywhere), priced at each family's current rate. Their context windows stay conservative — no fixed 1M grant, they keep the existing 200k base + >200k inference path, since the resolved version is unknown.
- **Unknown stays unpriced**: unparseable or malformed ids still return `None` → "—".
- **One judgment call to flag**: `claude-opus` (family word without a version) is kept **unpriced**, per your note about not pricing arbitrary strings that merely contain a family name — only the documented Claude Code aliases are exact-matched. There's a test asserting this; happy to flip it to alias behavior if you'd rather treat it as one.
- **Tests**: new coverage for `-latest`, all three aliases (price + window + `context_frac`), and the strict cases (`claude-opus`, `claude-opus-4-1-foo`, `claude-mythos-preview-extra`, embedded family names). `cargo test --locked`: 925 passed, 0 failed locally; clippy `-D warnings` and `fmt --check` clean.

Version-specific/legacy/preview pricing and the `mission_pricing` override are untouched.

RizRiyz:
@zengbo Thanks again for the alias work. I added a small follow-up in `7d4d4cd` for the newly documented Claude Code `fable` alias.

The change:

- recognizes only the exact `fable` alias
- uses Fable 5 pricing at $10 input, $50 output, and $1 cache read per MTok
- assigns its native 1M context window
- keeps `best`, `claude-fable`, and strings merely containing `fable` unpriced
- adds focused regression tests

This follows the current Claude Code model configuration and Anthropic pricing documentation.

Could you please take a look and let me know whether this behavior matches your understanding, especially treating `fable` as deterministically resolving to Fable 5 with a 1M window?

zengbo:
LGTM — verified the diff in `7d4d4cd`. Treating `fable` as deterministically Fable 5 is sound: unlike `opus`/`sonnet`/`haiku`, the family currently has exactly one model, so the "resolved version unknown → conservative window" concern doesn't apply, and the 1M window matches Fable 5's native window. The $10/$50/$1 rates match the Fable/Mythos 5 pricing already in this PR, and keeping `claude-fable`, `best`, and embedded strings unpriced is consistent with the strict-alias rule. One note for the future: if a second Fable version ever ships with different rates or window, this alias mapping will need revisiting — same caveat as the other family aliases' current-rate assumption. From my side the PR is ready to merge.

PR #143, opened 2026-08-26
feat(input): double-click a path/URL/word to copy it to the clipboard

sinardyas:
Double-clicking a path, URL, or word in a terminal pane copies it to the clipboard and highlights it.

## What

- Double-click in a terminal pane copies and highlights the token under the cursor.
- Paths and URLs (detected with `links::link_at`) copy as the full raw token: `src/main.rs:42:7` verbatim, soft-wrapped paths rejoined. Any other token falls back to the whitespace-delimited word; whitespace or empty cells copy nothing.
- Copying uses the pure `link_at`, not `link_at_screen`, so it skips the on-disk check: a `pwd` directory or a not-yet-created path still copies.
- A double-click is a second left press within 400ms on the same cell (±1), detected after the mouse-forward gate so it never steals a click from a mouse-tracking pane app. Reporting-off panes copy on a plain double-click; reporting-on panes (vim, tmux, agents) copy on Shift+double-click.
- Ctrl+click still opens links. A double-click's release keeps the highlight instead of re-copying; dragging afterwards reverts to an ordinary selection.
- Reuses the existing `pending_clipboard` and "Copied" toast; the highlight is a transient screen-coordinate `Selection`.
- Scope: terminal panes only. File-browser and diff views keep their existing selection behavior.

## Why

Terminals had no double-click behavior: crossterm emits no native double-click and the code never synthesized one, so a double-click only blinked the clicked cell and copied nothing. There was no quick way to grab a printed path such as `pwd` output. This adds the gesture on top of the existing link detection and clipboard paths.

## Verification

- [x] `cargo test --locked`
- [ ] `cargo clippy --all-targets -- -D warnings`

Run so far: `cargo test` (943 passing, including a new `word_at_grid` test) and `cargo clippy` (clean).

## Notes

- Single-character tokens copy but may not highlight (`Selection::has_range` treats a 1-cell selection as empty); rare for paths.
- Out of scope: triple-click line select, punctuation-aware word boundaries, and a config toggle (always on).
- Rebased on current `main` (v0.13.1) and reconciled with the recent Unicode-selection and scroll-stable-selection changes.
- One pre-existing flaky git test (`status_selection_tracks_the_exact_file_for_diff_activation`, "Bad file descriptor" under parallelism) is unrelated; it passes in isolation.

## Summary by CodeRabbit

- **New Features**
  - Double-click text in panes to copy and highlight complete URLs, file paths, or words.
  - Detection is limited to repeated clicks within the same pane.
  - Unicode paths, emoji, combining characters, and wide characters are preserved correctly.
  - Clicking either cell of a wide character selects the complete word.
  - Dragging after a double-click returns to standard text selection.

- **Bug Fixes**
  - Corrected title-case validation for leading minor words.

### DIFF
--- a/src/app/input.rs
+++ b/src/app/input.rs
@@ -97,9 +97,11 @@ const DOUBLE_CLICK: Duration = Duration::from_millis(400);
 type CellSpan = (u16, u16, u16);
 
 /// The whitespace-delimited word covering grid cell (`col`, `row`), and the one
-/// span it occupies. `None` on a whitespace or out-of-range cell. Char indices
-/// are the columns, matching how [`crate::links::link_at`] and the grid renderer
-/// address cells.
+/// span it occupies. `None` on a whitespace or out-of-range cell. Wide-cell
+/// continuation markers participate in the word boundary and are removed from
+/// the copied text, so either cell of `你` selects the same complete word. Char
+/// indices are the columns, matching how [`crate::links::link_at`] and the grid
+/// renderer address cells.
 fn word_at_grid(rows: &[String], col: u16, row: u16) -> Option<(String, Vec<CellSpan>)> {
     let line = rows.get(row as usize)?;
     let chars: Vec<char> = line.chars().collect();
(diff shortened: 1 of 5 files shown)


RizRiyz:
@sinardyas Awesome! The feature fits Luvus well, but two issues need fixing before merge:

- Double-click detection should only be armed inside terminal pane content and tied to the same pane. Currently, a quick title click followed by a nearby body click can be treated as a double-click, preventing the pane from receiving focus. This is causing the Linux CI failure.
- Token lookup must use terminal cell coordinates. visible_rows() omits wide-character spacer cells, so CJK or emoji before a token currently produces incorrect copying or highlighting.

Also add focused tests for pane title-to-body clicks, double-click copying, and wide-character text. GG!

PR #147, opened 2026-08-26
fix(server): force-stop when the named pipe no longer answers

### LINKED ISSUE #144: Windows: detached server stays up but stops answering, so reattach / ping / stop hang
**What happened?**
Closing the TUI is supposed to leave `luvus.exe server` running so the next `luvus` reattaches (tmux-style). That part is working as designed and should not change.

The bug is that the leftover server can keep the named pipe open and stop answering. After that, reattach fails:

- `luvus` (no args) does not show a usable TUI
- `luvus ping` can block for minutes
- `luvus server stop` can also block
- `luvus server status` reports the server is running but did not answer

Expected: a detached server remains attachable. If the control plane is dead, `ping` / `status` / `stop` / a new `luvus` return an error instead of hanging forever. This is not a request to stop the server when the TUI window closes, and not a request to kill a healthy session on a timeout.

**Steps to reproduce**
1. Run latest `luvus 0.13.1` in Windows Terminal.
2. Close the TUI (or otherwise detach). The background `luvus.exe server` stays, as designed.
3. Later, run `luvus` again to reattach, or run `luvus ping` / `luvus server status`.
4. Observe: the server process is still there, but status says it did not answer, and ping / reattach / stop may hang.

**Environment**
- OS + version: Windows 11 Pro 25H2 (10.0.26200)
- Terminal app + version: Windows Terminal 1.24.11911.0
- Luvus version and server status (`luvus server status`):

```text
luvus 0.13.1
Error: luvus 服务器正在运行但未响应: response read failed
```

```text
luvus 0.13.1

  ✓ core    多路复用器（窗格 · 标签页 · 智能体）无需外部工具
  ✓ git   git version 2.53.0.windows
  ✓ gh    gh version 2.92.0
  ✓ ssh   OpenSSH_10.2p1
  ✓ curl  curl 8.21.0 (Windows)
  · keys    请在 luvus 窗格外运行 `luvus doctor` 以测试终端

一切就绪。✓
```

**Anything else?**
While hung, three processes were seen at once: `luvus.exe server`, `luvus.exe ping`, `luvus.exe server stop` (no window). After killing only the stuck server, a fresh attach + ping worked, then the mute state returned later the same day (`session.json` still updating until it stopped answering).

Related: #116 bounded lifecycle *reads* (macOS PTY stall). That lets `status` error instead of hanging, but Windows `connect` and `ping` / TUI attach can still block. #47 is orphaned-server RSS on macOS, different bug.

Out of scope: stopping the server on TUI close; treating a live attachable session as dead because a timer fired.

Doeye1997:
A detached Windows server that still owns the named pipe but no longer handles requests no longer wedges `ping`, `server stop`, or reattach.

Closes #144.

## What

- Bound named-pipe `connect` so a listening-but-busy pipe cannot pin the CLI forever (`WaitNamedPipe` + timeout).
- `server stop` / attach reclaim: if the control plane does not answer, stop the pipe-owner PID (or `server.pid`) instead of hanging.
- Does not stop a healthy server when the TUI window closes.

## Why

#144: after detach, `luvus.exe server` can keep the pipe open and stop answering. `ping` / `stop` / a new `luvus` then hang. Expected: fail or reclaim, not wait forever.

This is the hang path. False `response read failed` after a successful write (`ERROR_PIPE_BUSY` / `PIPE_NOWAIT`) is #146, not this PR.

## Verification

- [x] Focused IPC/PID unit tests (`named_pipe`, `connect_timeout`, `pipe_nowait`, PeekNamedPipe deadline reader, start-marker, current-pid force-stop guard)
- [ ] `cargo clippy --all-targets -- -D warnings` (fails on pre-existing `src/ui/menu.rs` `nonminimal_bool` on `main`; that file is not in this PR)
- [ ] Manual: `server stop` against a mute Windows named-pipe server returns instead of hanging

## Notes

#146 is now on `main`. This rebase kept the complete PeekNamedPipe implementation (`set_recv_timeout` Windows path + `recv_has_data` + `api.rs` deadline reader). It did not reintroduce the incomplete deadline half-set.

This PR still owns `connect_timeout`, `endpoint_exists`, `server_pid`, force-stop, and recycle-waits-for-both-endpoints. Unrelated PI `--session` restore is not in `src/persist.rs`. `ServerPidFile` stays for force-stop.

Rebased onto `main`. Head `6bd37c1`. Force-with-lease pushed to this same branch.

### DIFF
--- a/src/ipc/transport.rs
+++ b/src/ipc/transport.rs
@@ -151,16 +151,8 @@ impl Conn {
         match self.0.set_send_timeout(Some(timeout)) {
             Ok(()) => Ok(TimeoutMode::Kernel),
             Err(error) if error.kind() == io::ErrorKind::Unsupported => {
-                #[cfg(windows)]
-                {
-                    let _ = timeout;
-                    Ok(TimeoutMode::Nonblocking)
-                }
-                #[cfg(not(windows))]
-                {
-                    self.0.set_nonblocking(true)?;
-                    Ok(TimeoutMode::Nonblocking)
-                }
+                self.0.set_nonblocking(true)?;
+                Ok(TimeoutMode::Nonblocking)
             }
             Err(error) => Err(error),
         }
--- a/src/ipc/transport.rs
+++ b/src/ipc/transport.rs
@@ -202,6 +202,47 @@ impl Conn {
             _ => Err(error),
         }
     }
+
+    /// PID of the process that owns the listening endpoint.
+    ///
+    /// Used by `server stop` when the app loop no longer answers: the pipe or
+    /// socket can still accept connections while requests hang forever.
+    pub fn server_pid(&self) -> io::Result<u32> {
+        #[cfg(windows)]
+        {
+            let Stream::NamedPipe(pipe) = &*self.0;
+            pipe.inner().server_process_id()
+        }
+        #[cfg(any(target_os = "linux", target_os = "android"))]
+        {
+            use std::mem::{size_of, zeroed};
+            use std::os::fd::AsRawFd;
+
+            let Stream::UdSocket(socket) = &*self.0;
+            let mut credentials: libc::ucred = unsafe { zeroed() };
+            let mut len = size_of::<libc::ucred>() as libc::socklen_t;
+            let result = unsafe {
+                libc::getsockopt(
+                    socket.inner().as_raw_fd(),
+                    libc::SOL_SOCKET,
+                    libc::SO_PEERCRED,
+                    (&raw mut credentials).cast(),
+                    &raw mut len,
+                )
+            };
+            if result != 0 {
+                return Err(io::Error::last_os_error());
+            }
+            Ok(credentials.pid as u32)
+        }
+        #[cfg(not(any(windows, target_os = "linux", target_os = "android")))]
+        {
+            Err(io::Error::new(
+                io::ErrorKind::Unsupported,
+                "server pid is not available on this transport",
+            ))
+        }
+    }
 }
 
 #[cfg(unix)]
@@ -431,6 +472,47 @@ pub(crate) fn discovery_address(path: &Path) -> String {
     }
 }
 
+/// Connect, but do not block the caller forever.
+///
+/// Windows named-pipe `connect` waits indefinitely when every instance is busy.
+/// A helper thread cannot fix that: `main` still waits for the blocked
+/// `connect` thread, so `ping` / `server stop` appear hung after the timeout.
+/// Wait on the pipe with a kernel timeout, then connect.
+pub fn connect_timeout(path: &Path, timeout: Duration) -> io::Result<Conn> {
+    #[cfg(windows)]
+    {
+        wait_for_named_pipe(path, timeout)?;
+        connect(path)
+    }
+    #[cfg(not(windows))]
+    {
+        let _ = timeout;
+        connect(path)
+    }
+}
+
+#[cfg(windows)]
+fn wait_for_named_pipe(path: &Path, timeout: Duration) -> io::Result<()> {
+    use windows_sys::Win32::System::Pipes::WaitNamedPipeW;
+
+    let name = format!(r"\\.\pipe\{}", pipe_id(path));
+    let wide: Vec<u16> = name.encode_utf16().chain(Some(0)).collect();
+    let ms = u32::try_from(timeout.as_millis()).unwrap_or(u32::MAX);
+    let ok = unsafe { WaitNamedPipeW(wide.as_ptr(), ms) };
+    if ok != 0 {
+        return Ok(());
+    }
+    let error = io::Error::last_os_error();
+    // ERROR_SEM_TIMEOUT (121): the wait elapsed with no free instance.
+    if error.raw_os_error() == Some(121) {
+        return Err(io::Error::new(
+            io::ErrorKind::TimedOut,
+            "connection timed out",
+        ));
+    }
+    Err(error)
+}
+
 /// Connect to a server socket identified by a per-session filesystem path.
 pub fn connect(path: &Path) -> io::Result<Conn> {
     #[cfg(windows)]
@@ -578,6 +660,40 @@ mod windows_security_tests {
             error.raw_os_error(),
             Some(windows_sys::Win32::Foundation::ERROR_PIPE_BUSY as i32),
             "expected ERROR_PIPE_BUSY after write, got {error}"
+        );
+    }
+
+    #[test]
+    fn client_sees_the_named_pipe_server_pid() {
+        let path = test_pipe("server-pid");
+        let listener = bind(&path).expect("bind owner-only named pipe");
+        let client_path = path.clone();
+        let expected = std::process::id();
+        let client = std::thread::spawn(move || {
+            let conn = connect(&client_path).expect("same-user client connects");
+            conn.server_pid().expect("named-pipe server pid")
+        });
+        let _server = listener.accept().expect("accept same-user client");
+        assert_eq!(client.join().unwrap(), expected);
+    }
+
+    #[test]
+    fn connect_timeout_fails_fast_when_the_pipe_is_absent() {
+        match connect_timeout(&test_pipe("missing"), Duration::from_secs(1)) {
+            Ok(_) => panic!("absent pipe must not connect"),
+            Err(err) => assert_ne!(err.kind(), io::ErrorKind::TimedOut),
+        }
+    }
+
+    #[test]
+    fn connect_timeout_returns_quickly_when_the_pipe_is_listening() {
+        let path = test_pipe("listening-no-accept");
+        let _listener = bind(&path).expect("bind owner-only named pipe");
+        let started = std::time::Instant::now();
+        let _ = connect_timeout(&path, Duration::from_millis(200));
+        assert!(
+            started.elapsed() < Duration::from_secs(2),
+            "connect_timeout must not pin the process on a listening pipe"
         );
     }
 }
--- a/src/ipc/transport.rs
+++ b/src/ipc/transport.rs
@@ -773,14 +773,9 @@ mod windows_security_tests {
     #[test]
     fn connect_timeout_times_out_when_every_instance_is_busy() {
         let path = test_pipe("busy-timeout");
-        let listener = bind(&path).expect("bind owner-only named pipe");
-        let occupied_path = path.clone();
-        let occupied = std::thread::spawn(move || connect(&occupied_path));
-        let _peer = listener.accept().expect("accept occupying client");
-        let first = occupied
-            .join()
-            .expect("occupying connect")
-            .expect("occupying client");
+        let _listener = bind(&path).expect("bind owner-only named pipe");
+        let first = connect_timeout(&path, Duration::from_millis(200))
+            .expect("first client occupies the free instance");
         let started = std::time::Instant::now();
         let result = connect_timeout(&path, Duration::from_millis(200));
         assert!(
@@ -789,10 +784,8 @@ mod windows_security_tests {
         );
         match result {
             Err(error) if error.kind() == io::ErrorKind::TimedOut => {}
-            Ok(_) => {
-                // Listener may have created another instance; still must be fast.
-            }
-            Err(error) => panic!("unexpected connect_timeout error: {error}"),
+            Ok(_) => panic!("busy pipe must time out, got a connection"),
+            Err(error) => panic!("busy pipe must time out, got {error}"),
         }
         drop(first);
     }
--- a/src/ipc/transport.rs
+++ b/src/ipc/transport.rs
@@ -9,6 +9,8 @@ use std::io::{self, Read, Write};
 use std::path::Path;
 use std::sync::Arc;
 use std::time::Duration;
+#[cfg(windows)]
+use std::time::Instant;
 
 use fs2::FileExt;
 use interprocess::local_socket::prelude::*;
@@ -478,15 +480,13 @@ pub(crate) fn discovery_address(path: &Path) -> String {
 
 /// Connect, but do not block the caller forever.
 ///
-/// Windows named-pipe `connect` waits indefinitely when every instance is busy.
-/// A helper thread cannot fix that: `main` still waits for the blocked
-/// `connect` thread, so `ping` / `server stop` appear hung after the timeout.
-/// Wait on the pipe with a kernel timeout, then connect.
+/// Windows named-pipe `CreateFile` waits indefinitely when every instance is
+/// busy. Loop `CreateFile` / `WaitNamedPipe` against the remaining deadline
+/// instead of calling unbounded `connect()` after one successful wait.
 pub fn connect_timeout(path: &Path, timeout: Duration) -> io::Result<Conn> {
     #[cfg(windows)]
     {
-        wait_for_named_pipe(path, timeout)?;
-        connect(path)
+        connect_named_pipe_timeout(path, timeout)
     }
     #[cfg(not(windows))]
     {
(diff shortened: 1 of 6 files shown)


src/platform.rs
```diff
@@ -316,6 +316,63 @@ pub fn process_belongs_to_current_user(pid: u32) -> bool {
...
+    }
+    #[cfg(target_os = "linux")]
+    {
+        return std::fs::read_to_string(format!("/proc/{pid}/comm"))
+            .is_ok_and(|comm| comm.trim() == "luvus");
+    }
+    #[cfg(all(unix, not(target_os = "linux")))]
+    {
+        return true;
+    }
```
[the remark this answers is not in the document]
Doeye1997: Fixed: macOS/BSD now require process_tree command basename `luvus` before force-stop. `server.pid` also stores a start marker so a reused PID is ignored.

src/main.rs
```diff
@@ -776,16 +784,36 @@ fn server_status() -> Result<()> {
...
             // or Windows may observe the named pipe closing before the final
             // stop acknowledgement is readable. A completed shutdown is still
-            // success; a live, unresponsive server keeps the original error.
+            // success. A live, unresponsive server is force-killed.
+            if force_stop_unresponsive(pid, &client_socket).is_ok() {
+                return Ok(true);
+            }
             if wait_for_shutdown(&client_socket).is_ok() {
                 return Ok(true);
             }
```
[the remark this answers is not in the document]
Doeye1997: Fixed differently: on a `server.stop` reply error we wait for the socket to vanish first. Force-kill only if the server is still there and the PID is a stoppable Luvus process.

src/main.rs
```diff
@@ -440,13 +421,45 @@ fn autodetect_and_attach() -> Result<()> {
...
+            wait_for_socket(sock)
+        }
+    }
+}
+
+fn recycle_unresponsive_server(sock: &Path) -> Result<()> {
+    send_server_stop()?;
+    spawn_server()?;
+    wait_for_socket(sock)
+}
```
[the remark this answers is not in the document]
Doeye1997: Fixed differently: if connect succeeds we no longer recycle on a version-read error. Recycle stays only for connect timeout (pipe exists but nobody accepts). No extra retry helper.

src/persist.rs
```diff
@@ -429,6 +429,33 @@ pub fn session_dir() -> PathBuf {
...
+            .parse()
+            .ok()
+    }
+}
+
+impl Drop for ServerPidFile {
+    fn drop(&mut self) {
+        let _ = fs::remove_file(session_dir().join("server.pid"));
+    }
+}
```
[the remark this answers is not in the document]
Doeye1997: Fixed: `ServerPidFile` writes `pid start-marker` and `read` drops the record when the live marker does not match. macOS/BSD stoppable check is the process name, not `true`.

Doeye1997:
Addressed the follow-up on API timeout vs client socket: if the API pipe times out but the client socket still connects, `server stop` now returns an error instead of force-killing `server.pid`. Force-stop remains only when both sockets time out (or the client socket is absent after an API timeout).

Doeye1997:
Follow-up: if `server.stop` connects but the 1s acknowledgement times out and the client socket still accepts, stop now returns an error instead of force-killing. Force-stop remains only when the client socket is gone too.

Doeye1997:
Root cause for the recycle race: `recycle_unresponsive_server` spawned immediately after stop, while the old process could still own the API or client pipe. The replacement then saw `endpoint_exists` and exited; the old process died next — no server left. It now waits until both endpoints are gone before spawn. `wait_for_shutdown` uses the same both-endpoint check (TimedOut still counts as present).

RizRiyz:
@Doeye1997 PR #147 should remain independent from #146.

The current branch appears to contain part of #146’s Windows response deadline implementation in `src/ipc/transport.rs`, but not the matching reader changes in `src/ipc/api.rs`. This leaves the timeout fallback incomplete and can still allow a blocking read.

You need either:

- remove the #146-specific `set_recv_timeout` and `recv_has_data` changes from this PR, or
- rebase after #146 is merged and preserve the complete implementation.

Since the PR body says the false response error belongs to #146, removing that partial overlap is probably the cleaner option.

Also remove the unrelated PI session restoration changes from `src/persist.rs`, then rebase the branch and run the currently missing CI and Windows verification.

Doeye1997:
> PR #147 should remain independent from #146.

Done. Dropped the incomplete deadline half-set (`recv_has_data` / PeekNamedPipe, and the Windows `set_recv_timeout` fallback). Timeout helpers match `main`. `connect_timeout` and `server_pid` stay here. The PeekNamedPipe reader belongs in #146 with `api.rs`.

> Also remove the unrelated PI session restoration changes from `src/persist.rs`

Removed. `ServerPidFile` stays for force-stop.

Rebased onto `main`. Head `633480c`. GitHub CI is green, including Windows ConPTY. Have not run mute-pipe `server stop` by hand.

RizRiyz:
@Doeye1997 I think you still need to rebase and resolve the conflicts

Doeye1997:
Rebased `fix/unresponsive-server-stop` onto current `origin/main` (includes merged #146) and force-with-lease pushed to this same branch. Conflicts are resolved; this PR stays independent of #146's hang-path work.

**Kept (this PR)**
- Force-stop when both the API and client sockets time out (or the client socket is absent after an API timeout)
- Error instead of killing `server.pid` when the API times out but the client socket still connects
- Recycle waits until both endpoints are gone before spawn
- `connect_timeout`, `endpoint_exists`, `server_pid`, `ServerPidFile` (with start-marker stale-PID rejection)

**Kept from main / #146 (complete implementation)**
- `set_recv_timeout` Windows path that does **not** switch to `PIPE_NOWAIT`
- `recv_has_data` / `PeekNamedPipe`
- `src/ipc/api.rs` deadline reader

**Dropped (not reintroduced)**
- Incomplete deadline half-set that this branch used to carry without the matching reader
- Unrelated PI `--session` restore in `src/persist.rs`

Head is `6bd37c1`. `cargo fmt --all --check` passed. Focused Windows IPC/PID tests passed (named-pipe connect timeout, busy-pipe presence, server PID, `PIPE_NOWAIT`-after-write, PeekNamedPipe deadline reader, start-marker, current-pid force-stop guard). `cargo clippy --all-targets -- -D warnings` fails on pre-existing `src/ui/menu.rs` `nonminimal_bool` on `origin/main`; that file is not in this PR. Mute-pipe `server stop` still not run by hand against a live isolated server.

src/main.rs
```diff
@@ -451,13 +432,51 @@ fn autodetect_and_attach() -> Result<()> {
...
+                        "luvus v{binary} installed, but the running server is v{running} — \
+                         run `luvus server restart` to load it (your session is saved and restored)."
+                    );
+                    thread::sleep(Duration::from_millis(2000));
+                }
+                Ok(())
+            }
+            // Connected: the pipe accepted. A slow or false IO reply is not a
+            // mute server. #144: do not kill a healthy session on a timeout.
+            Err(_) => Ok(()),
```
Doeye1997: **[bug]** `ensure_server_ready` treats a successful client-pipe connect as proof the server is usable, then ignores a failed `server_version()` ping.

That is the #144 failure mode: accept threads stay alive (`start_client_listener` / `api::start_server` are dedicated threads), the app loop no longer handles requests. This function returns `Ok(())`, then `ipc::client::run` uses unbounded `transport::connect` and waits forever for frames, so reattach still wedges.

Recycle (using the existing `is_stoppable_luvus_pid` guard) or refuse to attach when the control ping times out. Do not treat connect success as liveness. A 1s ping timeout on a healthy server is a separate question from attaching to a mute one.

src/ipc/transport.rs
```diff
@@ -431,6 +476,72 @@ pub(crate) fn discovery_address(path: &Path) -> String {
...
+///
+/// Windows named-pipe `connect` waits indefinitely when every instance is busy.
+/// A helper thread cannot fix that: `main` still waits for the blocked
+/// `connect` thread, so `ping` / `server stop` appear hung after the timeout.
+/// Wait on the pipe with a kernel timeout, then connect.
+pub fn connect_timeout(path: &Path, timeout: Duration) -> io::Result<Conn> {
+    #[cfg(windows)]
+    {
+        wait_for_named_pipe(path, timeout)?;
+        connect(path)
```
Doeye1997: **[bug]** Windows `connect_timeout` still calls unbounded `connect()` after `WaitNamedPipeW` succeeds.

The comment on this function is that named-pipe `connect` waits indefinitely when every instance is busy. `WaitNamedPipe` + `CreateFile` is a known race: another client can take the instance between the wait and `connect()`, and `Stream::connect` then blocks with no timeout — the original hang.

The usual pattern is to loop `CreateFile` / `WaitNamedPipe` against the remaining deadline, and return `TimedOut` on `ERROR_PIPE_BUSY` after the budget is gone. Do not call unbounded `connect()` here.

`connect_timeout_returns_quickly_when_the_pipe_is_listening` does not catch this: it ignores the result and only asserts elapsed < 2s. If `WaitNamedPipe` times out, `connect()` is never called.

src/main.rs
```diff
@@ -850,19 +866,54 @@ fn print_detached_status(context: i18n::cli::Context) {
...
-            // success; a live, unresponsive server keeps the original error.
+            // Stop may have already taken effect: the pipe closed before the
+            // acknowledgement was readable. Wait for the socket to vanish
+            // before force-killing, so a healthy shutdown is not interrupted.
             if wait_for_shutdown(&client_socket).is_ok() {
                 return Ok(true);
             }
+            // Client socket still accepts: server is reachable. The stop
+            // request timed out; do not kill the session.
+            if ipc::transport::connect_timeout(&client_socket, SERVER_CONTROL_TIMEOUT).is_ok() {
```
Doeye1997: **[bug]** After a timed-out `server.stop` reply, a client socket that still accepts is treated as reachable and force-stop is skipped.

`start_client_listener` runs on its own thread, so a dead app loop still accepts. This path first spends `wait_for_shutdown` polling endpoints that are still live (seconds), then returns an error without reclaiming. Combined with `ensure_server_ready` swallowing ping failures, `server stop` cannot reclaim and `luvus` cannot reattach — the hung process keeps the pipes.

The PR text allows fail-or-reclaim instead of hanging forever. Failing is only useful if attach also fails closed. As written, stop errors after a long poll and attach still hangs.

Force-stop when stop is not acknowledged and the PID passes `is_stoppable_luvus_pid`, or fail fast without the shutdown poll if you are intentionally not reclaiming this state.

src/platform.rs
```diff
@@ -316,6 +316,72 @@ pub fn process_belongs_to_current_user(pid: u32) -> bool {
...
+            std::process::Command::new("taskkill")
+                .args(["/PID", &pid.to_string(), "/T", "/F"])
+                .stdout(std::process::Stdio::null())
+                .stderr(std::process::Stdio::null()),
+        )
+        .status();
+    }
+    #[cfg(unix)]
+    unsafe {
+        libc::kill(pid as libc::pid_t, libc::SIGKILL);
```
Doeye1997: **[bug]** The comment says this ends `pid` and its children, but the Unix arm only `SIGKILL`s that one pid.

Windows uses `taskkill /T /F`. Unix servers are started with `setsid()`, so pane PTYs become orphans after this kill. That contradicts the product rule that stopping the server ends its live PTYs.

Kill the process group (or equivalent tree) on Unix after the stoppable check, and do not ignore `kill` failure before `wait_for_shutdown`.

Doeye1997 (commented):
## Requesting changes

This PR bounds some CLI control-plane waits and adds a PID force-stop path, but it does not unwedge the failure mode stated in the PR and in `219ed7a`: a detached server whose **accept threads still answer** while the app loop no longer handles `server.stop` or client frames.

`api::start_server` and `start_client_listener` run on dedicated threads, so a successful `connect_timeout` is not evidence the server can handle requests. After the later commits that refuse to kill on a timed-out stop reply, both `server stop` and attach leave that process in place. Attach then calls unbounded `ipc::client::run` -> `transport::connect` and waits forever for frames. That is still the #144 reattach hang.

You cannot both (a) never reclaim when a pipe still accepts and (b) unwedge reattach for a mute app loop. Current code chooses (a) and leaves (b) broken. `ping`/`stop` now fail instead of hanging forever, which is an improvement, but reattach still wedges and `server stop` cannot reclaim the process.

### What looks solid

- `ServerPidFile` start-marker reuse check
- `is_stoppable_luvus_pid` rejects pid 0 and self
- Unix `connect_timeout` uses `ConnectWaitMode::Timeout`
- Recycle waits for both endpoints before spawn
- Windows `server_pid()` via the named-pipe server process id

### Blockers

1. `ensure_server_ready` swallows a failed `server_version()` ping after a successful client connect, then attach hangs.
2. Windows `connect_timeout` still calls unbounded `connect()` after `WaitNamedPipeW`.
3. `send_server_stop` skips force-stop when the client socket still accepts (the mute-app-loop case), after first polling live endpoints in `wait_for_shutdown`.
4. Unix `force_terminate` SIGKILLs only that pid; Windows uses `taskkill /T`. Orphaned pane PTYs survive a force-stop.

### Tests

No coverage of `ensure_server_ready`, `send_server_stop`, or `ServerPidFile` marker mismatch. The Windows listening-pipe `connect_timeout` test ignores the result and only checks elapsed time, so it cannot catch an unbounded `connect()` after a successful wait.

src/main.rs
Doeye1997: Addressed in 4118e2f.

`ensure_server_ready` no longer treats a successful client-pipe connect as liveness. A failed `server_version()` ping now calls `recycle_unresponsive_server`, which force-stops a stoppable Luvus pid and waits for both endpoints before spawn. Reattach will not fall through to unbounded `ipc::client::run` against a mute app loop.

src/ipc/transport.rs
Doeye1997: Addressed in 4118e2f.

Windows `connect_timeout` no longer calls unbounded `connect()` after `WaitNamedPipe`. It loops `CreateFileW` / `WaitNamedPipeW` against the remaining deadline and returns `TimedOut` on `ERROR_PIPE_BUSY` / `ERROR_SEM_TIMEOUT` when the budget is gone.

`connect_timeout_returns_quickly_when_the_pipe_is_listening` now asserts Ok or TimedOut. Added `connect_timeout_times_out_when_every_instance_is_busy` for the occupied-instance race.

src/main.rs
Doeye1997: Addressed in 4118e2f.

`send_server_stop` no longer skips force-stop when the client socket still accepts. An API timeout with a live client socket, or a missing `server.stop` ack with live endpoints, force-stops if the pid passes `is_stoppable_luvus_pid`. One short endpoint probe distinguishes an already-finished shutdown from a mute accept thread. Combined with the ping-failure recycle, attach fails closed or replaces the process instead of hanging.

src/platform.rs
Doeye1997: Addressed in 4118e2f.

Unix `force_terminate` now `SIGKILL`s the process group when `getpgid(pid) == pid` (the `setsid` server case), then walks `process_tree` so pane PTYs die with the server. Kill failure is returned unless the pid is already gone. Windows still uses `taskkill /T /F`.

Doeye1997 (commented):
Addressed in 4118e2f.

`ensure_server_ready` no longer treats a successful client-pipe connect as liveness. A failed `server_version()` ping now calls `recycle_unresponsive_server`, which force-stops a stoppable Luvus pid and waits for both endpoints before spawn. Reattach will not fall through to unbounded `ipc::client::run` against a mute app loop.

Doeye1997:
Addressing review 5043841648 in 4118e2f.

1. `ensure_server_ready`: failed control ping after connect now recycles (connect is not liveness).
2. Windows `connect_timeout`: `CreateFileW` / `WaitNamedPipeW` loop against remaining deadline; no unbounded `connect()`.
3. `send_server_stop`: force-stop when stop is not acked and a stoppable pid exists, even if the client socket still accepts.
4. Unix `force_terminate`: SIGKILL process group + `process_tree`; kill errors are not ignored.

Threads for those four comments are replied to and resolved.

src/main.rs
Doeye1997: Follow-up tests in 553e2db for the remaining coverage item:

- `send_server_stop_fails_closed_when_a_listener_never_replies`
- `ensure_server_ready_fails_closed_when_control_ping_never_replies`
- `connect_timeout_times_out_when_every_instance_is_busy` now requires `TimedOut` (occupies the free instance, then the second connect must time out)
- `ServerPidFile` mismatch already covered in 4118e2f

A mute listener with no stoppable foreign pid fail-closes instead of hanging. That is the attach-safe contract when reclaim cannot run.

Doeye1997:
Tests from review 5043841648 are in 553e2db:

- `ServerPidFile::read` -> `None` on mismatched start marker
- mute listener: `send_server_stop` / `ensure_server_ready` fail closed, no hang
- busy Windows pipe: `connect_timeout` returns `TimedOut` within the deadline

The inline tests comment (`discussion_r3874289257`) has no GraphQL thread node (`PRRT_kwDOTBelYM6c6ucy` does not resolve). Coverage is on this comment instead.

Doeye1997 (commented):
## Approving

Head `553e2db` addresses the four blockers from review `5043841648`.

- `ensure_server_ready` recycles when the client pipe accepts but `server_version()` fails, so reattach no longer proceeds into unbounded `ipc::client::run`.
- Windows `connect_timeout` loops `CreateFile` / `WaitNamedPipe` against the remaining deadline instead of calling unbounded `connect()` after one successful wait. Flags match interprocess `connect_without_waiting`.
- `send_server_stop` force-stops when `server.stop` is not acknowledged and the PID passes `is_stoppable_luvus_pid`. The long `wait_for_shutdown` poll before reclaim is gone.
- Unix `force_terminate` SIGKILLs the process group when the target is session leader, then walks `process_tree`, and returns the kill error if the pid is still stoppable.

Tests now cover mute-accept stop/attach fail-closed, busy-pipe timeout, pid-file marker mismatch, and Unix setsid kill.

Residual, not blocking: `--local` occupancy still uses `connect_timeout(...).is_ok()` (`src/main.rs` around the startup lock) while `ipc/server.rs` uses `endpoint_exists` (timeout == occupied). A busy/hung pipe is therefore not treated as an active server on the `--local` path. `FILE_FLAG_FIRST_PIPE_INSTANCE` should still fail the bind, so this is a worse error rather than a second server. Same helper as `server.rs` would make the occupancy rule one code path.

Doeye1997 (commented):
## Approving

Head `4dcf5cf` keeps the mute-accept reclaim path from `553e2db` and adds a second control-plane probe so a single 1s deadline no longer force-stops a reachable server.

- `ensure_server_ready` still recycles after connect+ping fail, but only after `retry_control_probe` (1s then 4s). Recycle passes `ConfirmedUnresponsive`, so attach does not spend a second confirmation ping or fall through to unbounded `ipc::client::run`.
- Client-pipe `TimedOut` now pings the API with the recovery deadline: a busy listener on a live app loop errors out (`retry attach`) instead of killing the session.
- `send_server_stop` uses the same confirmation before force-stop on the CLI path; a missing stop ack still reclaims when confirmation fails.
- Windows `is_stoppable_luvus_pid` uses ToolHelp image name instead of `process_tree` argv, so `luvus.exe server` is actually stoppable. macOS uses `proc_pidpath`. Unix test covers a luvus-named executable with arguments.

Mute-accept tests now inject the shorter timeouts. Residual, not blocking: `--local` occupancy still uses `connect_timeout(...).is_ok()` (`src/main.rs` around the startup lock) while `ipc/server.rs` uses `endpoint_exists` (timeout == occupied). CLI `server stop` of a busy-but-alive API also errors instead of retrying `server.stop` with the recovery timeout; that is fail-closed, not a hang.

Doeye1997 (commented):
## Approving

Head `8a5a316` is the already-reviewed recovery commit rebased onto current `main` (`6bdf9d0`, including shared process snapshots). The mute-accept reclaim path is unchanged.

- `ensure_server_ready` recycles only after `retry_control_probe` (1s then 4s) fails. Recycle passes `ConfirmedUnresponsive`, so attach does not fall through to unbounded `ipc::client::run`.
- Client-pipe `TimedOut` pings the API with the recovery deadline and errors out on a live app loop instead of killing the session.
- CLI `send_server_stop` confirms before force-stop; a missing stop ack still reclaims when confirmation fails.
- Windows `is_stoppable_luvus_pid` still uses ToolHelp image name (`ProcessSnapshot.names`), which remains the executable basename after the snapshot-sharing rebase. macOS still uses `proc_pidpath`.

Residual, not blocking: `--local` occupancy still uses `connect_timeout(...).is_ok()` (`src/main.rs` around the startup lock) while `ipc/server.rs` uses `endpoint_exists` (timeout == occupied). CLI `server stop` of a busy-but-alive API still errors instead of retrying `server.stop` with the recovery timeout; that is fail-closed, not a hang.

Doeye1997 (commented):
## Approving

Head `460870b` closes the `--local` occupancy residual from reviews on `553e2db` / `4dcf5cf` / `8a5a316`.

`--local` now uses `endpoint_exists` (timeout == occupied), matching `ipc/server.rs`. A busy or hung pipe is treated as an active server instead of an empty slot, so `--local` will not try to bind over a mute accept thread.

The mute-accept reclaim path is unchanged. Residual, not blocking: CLI `server stop` of a busy-but-alive API still errors instead of retrying `server.stop` with the recovery timeout; that is fail-closed, not a hang.

PR #150, opened 2026-08-27
fix(detect): grok reads Working while a subagent runs

riclib:
## What

Grok already has a per-agent working hint for its `Ctrl+c:cancel` footer. While a subagent is running, Grok replaces that footer entirely with:

```
◎ 1 subagent still running · send a message to interrupt
```

so the existing hint stops matching, and the pane reads `Idle` for however long the subagent takes — minutes, in practice.

## How it showed up

On 0.13.1, a Grok pane nine minutes into a subagent reported `Idle`. `luvus agent explain` was unambiguous that nothing had matched:

```json
"state_evidence": {
  "source": "no_positive_state_evidence",
  "confidence": "none",
  "rule_priority": null,
  "rule_region": null
}
```

So `Idle` was the fallback rather than a finding. With the rule, the same pane gives `source: manifest_rule`, `confidence: high`, `rule_priority: 105`.

I found this while building a bar widget that renders `agent list`, where "idle" and "we could not tell" are indistinguishable to the reader — but the gap is in detection, not in anything downstream.

## Why per-agent rather than `WORKING_HINTS`

Following the comment on the rule directly above it: Grok's punctuation is its own, which is why `ctrl+c:cancel` is a per-agent hint instead of a generic one. Same applies here, and it keeps the string away from the other fourteen agents. The underlying reasoning is the one `WORKING_HINTS` is built on though — an invitation to interrupt only exists while there is something to interrupt.

## Checks

- `cargo fmt --all --check` — clean
- `cargo test --locked` — 955 passed, 0 failed

Adds `detect::tests::grok_running_a_subagent_is_working`, built from the real pane text.

## Caveat

This covers the subagent footer specifically. I have not observed what Grok prints during an ordinary turn that shows neither footer, so there may be a further gap I could not test for.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Grok activity detection when a subagent is running.
  * Grok panes now correctly show as working when displaying the “send a message to interrupt” footer.
  * Existing cancellation prompts continue to be recognized.

### DIFF
--- a/src/detect.rs
+++ b/src/detect.rs
@@ -589,6 +589,19 @@ fn builtin_rules() -> Vec<Rule> {
             Region::Screen,
             vec![any(&["ctrl+c:cancel"])],
         ),
+        // While a subagent is running, Grok replaces that footer entirely with
+        // "1 subagent still running - send a message to interrupt", so the rule
+        // above stops matching and the pane reads Idle for however long the
+        // subagent takes - minutes, in practice. Same reasoning as WORKING_HINTS:
+        // an invitation to interrupt only exists while there is something to
+        // interrupt.
+        per(
+            "grok",
+            State::Working,
+            105,
+            Region::Screen,
+            vec![any(&["send a message to interrupt"])],
+        ),
     ]
 }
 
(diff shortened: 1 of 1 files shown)


RizRiyz:
@riclib great fix, you need to update both the PR title and commit subject to Luvus’s Conventional Commit format from our [contribution guide](https://github.com/RizRiyz/luvus?tab=contributing-ov-file#commits) :

The current `detect:` prefix is not a Conventional Commit type. `detect` should be the scope under `fix`.

riclib:
Thanks — fixed. PR title and commit subject are both `fix(detect): grok reads Working while a subagent runs` now; amended and force-pushed as `a1d9d15`, no content change.

RizRiyz:
Thank you @riclib approved!

PR #184, opened 2026-08-29
fix(orch): isolate ledger path during tests

### LINKED ISSUE #163: Flaky: wheel_scroll_moves_the_cursor_clamped picks up an orch task from another test
## What I saw

While running the suite on a feature branch, `app::board::tests::wheel_scroll_moves_the_cursor_clamped` failed intermittently — roughly 2 times in 19 full-suite runs:

```
thread 'app::board::tests::wheel_scroll_moves_the_cursor_clamped' panicked at src/app/board.rs:894:9:
assertion `left == right` failed
  left: 3
 right: 2
```

The test creates three tasks and asserts the cursor clamps at index 2. Getting 3 means the app had **four** tasks — so one arrived from outside the test.

## Why I think it happens

`OrchState` is loaded from disk at `App::new` (`src/app/mod.rs:1877` → `OrchState::load`), and the path is ambient:

```rust
fn orch_path() -> PathBuf {
    crate::persist::session_dir().join("orch.json")   // src/orch/mod.rs:449
}
```

`session_dir()` resolves through `LUVUS_HOME`, which `persist::test_env` sets with `std::env::set_var` — process-global. `test_env` serializes its callers with `TEST_ENV_LOCK`, but that only covers tests that opt in.

Meanwhile a lot of app-level orch mutation persists eagerly — `self.orch.save()` appears 11 times in `src/app/board.rs` and several more in `src/app/dispatch.rs`.

So a test that drives one of those code paths *without* taking `test_env` writes `orch.json` into whichever test's private `LUVUS_HOME` happens to be active at that moment. The next `App::new()` under that home loads the stray task, and a test that counts its own tasks sees one too many. Timing-dependent, which fits the intermittency.

## What I have and have not established

- The mechanism above is read from the code; I have not instrumented it.
- **I have not identified the specific polluting test.** The obvious suspect in `board.rs`, `agent_launch_line_is_one_quoted_line_with_the_contract`, builds an `OrchState::default()` purely in memory and never saves, so it is innocent.
- I could not reproduce it on pristine `main` — 0 failures in 30 full-suite runs at 5e8ad34, versus ~2 in 19 on a branch that adds about ten unrelated tests. I read that as the added tests shifting the interleaving rather than causing it, but I want to be straight that it is inference: I have not demonstrated a failure on untouched `main`.

If it is worth pinning down properly, the cheap next step is a temporary assert (or a `dbg!`) on `OrchState::load()` returning non-empty inside `test_env`, which should name the writer quickly.

## Possible fixes

Smallest: have the orch-touching tests that currently skip `test_env` take it, so every test that can reach `orch.save()` is inside the lock.

Structural, if you would rather close it for good: let the orch path be injected rather than read from ambient process state, so test isolation does not depend on every test remembering to opt in.

Happy to send either — say which you prefer, or close this if it is known.

## Environment

Linux, `cargo test --locked`, at 5e8ad34 and on a branch off it. Note the suite also has a separate, fully deterministic failure when run from a **git worktree**: `app::dispatch::tests::agent_list_labels_a_pane_with_its_node_name` asserts `worktree == false`, which cannot hold in a linked worktree (`.git` is a file there). That one is unrelated to this issue and reproduces every run — mentioning it only so it is not confused with the flake above. Glad to file it separately if useful.

Doeye1997:
Test isolation no longer lets one orch test leak tasks into another via process-global `$LUVUS_HOME`.

Fixes #163

## What

- `OrchState` now stores the `orch.json` path captured at load. Later `save()` writes that path instead of re-reading the ambient session dir.
- In the test binary, each `load()` gets a unique temp file, so a test that never takes `test_env` cannot read or write whichever home another test currently holds.
- `OrchState::default()` stays in-memory: `save()` is a no-op without a captured path.
- Added tests for explicit-path roundtrip, per-load isolation, captured-path save, and default no-op.

## Why

`app::board::tests::wheel_scroll_moves_the_cursor_clamped` sometimes saw four tasks instead of the three it created. `OrchState::load()`/`save()` used `session_dir()`, which follows process-global `$LUVUS_HOME`. `test_env` serializes only callers that opt in; a parallel test that mutates orch without the lock can write into the home currently held by another test. The next `App::new()` under that home loads the stray task.

## Verification

- [x] `cargo test --locked orch::`
- [x] `cargo test app::board::`
- [ ] `cargo test --locked` (full suite not completed here: Windows run hung on existing PTY/`test_env` tests after the focused checks passed)
- [x] `cargo test --locked diff_api_status_scan_is_off_loop` (passes in isolation)

## Notes

- Production path is unchanged: `load()` still reads `<session_dir>/orch.json` and `save()` now pins that captured path for the `App` lifetime.
- `app::dispatch::tests::agent_list_labels_a_pane_with_its_node_name` still fails in a linked git worktree (`worktree == false` cannot hold). That is the separate issue called out in #163, not this race.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured orchestration state saves to the location captured when it was loaded, even if the session directory changes later.
  * Prevented unintended file writes for in-memory orchestration state.
  * Improved isolation between test session data.

* **New Features**
  * Added support for loading orchestration state from an explicitly specified path.

### DIFF
--- a/src/orch/mod.rs
+++ b/src/orch/mod.rs
@@ -114,6 +114,10 @@ pub struct OrchState {
     next_task: u64,
     #[serde(default)]
     next_lease: u64,
+    /// Captured at load so later `save()` does not follow a process-global
+    /// session-dir change. Empty only on `Default` (in-memory, no persistence).
+    #[serde(skip)]
+    persist_path: PathBuf,
 }
 
 /// Why a mutation was rejected — carried to the API as a `(code, message)` error.
@@ -423,16 +427,26 @@ impl OrchState {
     // ── persistence (separate file; never touches session.json) ────────────
 
     pub fn load() -> OrchState {
-        match std::fs::read_to_string(orch_path()) {
+        Self::load_from(persist_path())
+    }
+
+    pub fn load_from(path: PathBuf) -> OrchState {
+        let mut state = match std::fs::read_to_string(&path) {
             Ok(s) => serde_json::from_str(&s).unwrap_or_default(),
             Err(_) => OrchState::default(),
-        }
+        };
+        state.persist_path = path;
+        state
     }
 
     /// Atomic save (temp + rename), best-effort — a failed write never breaks
     /// the app; the ledger is a convenience layer, not core session state.
+    /// `Default` (empty `persist_path`) is in-memory only and does not write.
     pub fn save(&self) {
-        let path = orch_path();
+        let path = &self.persist_path;
+        if path.as_os_str().is_empty() {
+            return;
+        }
         if let Some(dir) = path.parent() {
             let _ = std::fs::create_dir_all(dir);
         }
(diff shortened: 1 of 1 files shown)


RizRiyz:
@Doeye1997 Thanks for the PR! The orchestration layer is currently being refactored from the ground up, so feel free to pick up another issue or feature in luvus in the meantime.

RizRiyz:
Moved to #189 and your contribution still be recorded there, thanks @Doeye1997

Issue #170, opened 2026-08-28
Shift+click gestures may never reach Luvus: terminals claim Shift to bypass mouse reporting


riclib:
## What happens

On Alacritty 0.17.0 with a default mouse configuration, holding Shift while
clicking a FILES row does nothing at all. The mouse pointer changes the instant
Shift goes down: the terminal has taken the mouse back for its own selection and
stopped forwarding events to the application. Luvus never receives the click, so
the gesture is a no-op both on the file already previewed and on any other file.

## The app side is fine

`handle_mouse` reads the modifier correctly — `src/app/input.rs:1887` for FILES
rows, `src/app/input.rs:1872` for DIFF rows — and the resulting behaviour is
covered by unit tests that pass. Nothing is wrong downstream of the event. The
event does not arrive.

## Where it bites

- **FILES**: Shift+click is documented as the permanent read-only pane
  (`website/src/content/docs/docs/guides/files.mdx:51`, "the fixed *keep this
  one open* gesture").
- **DIFF rows**: same modifier, same fate.

Holding Shift to bypass mouse reporting is the xterm convention and Alacritty
follows it, so this is not an Alacritty bug — it is a gesture the terminal layer
is entitled to claim. Worth checking each terminal Luvus targets: the ones that
do this make every Shift+mouse gesture unreachable. Luvus's own shortcut help
already describes `shift+drag` as "select inside a mouse-aware app", which is
the same mechanism described from the other side.

## Why it matters more now

Found while click-testing #164. With `File click behavior: Preview` as the new
default, a plain click gives a recyclable preview and Shift+click is the
advertised way to make one stick. On a terminal that eats the modifier there is
then **no mouse gesture that produces a permanent pane**. Nothing becomes
unreachable — right-click → `Open (Read-Only)` still opens a permanent tab — but
the documented route is the one that does not work, and it is documented without
a caveat.

## Options

- **An alternate modifier.** Ctrl and Alt are forwarded normally by terminals
  that claim Shift.
- **An `Open in Pane` row in the FILES context menu.** The DIFF menu already has
  one, so this is an existing pattern rather than a new concept.
- **Document the caveat** and leave the gesture to terminals that forward it.

## Reproducing

1. Alacritty 0.17.0, default mouse configuration.
2. Open the FILES dock and Shift+click any row.
3. Nothing happens. The pointer changing shape as Shift goes down is the tell:
   that is the terminal switching from application mouse mode to its own
   selection.

[...]

PR #186, opened 2026-08-29
feat(files): open Ctrl+clicked paths where File click behavior says

### LINKED ISSUE #185: FILES: Ctrl+click on a printed path should respect File click behavior
**What problem would this solve?**

Since #164, **File click behavior** (Settings → General) decides where a file lands when you click it in the FILES tree — a reused preview pane beside you by default, or a tab. But a `Ctrl`+click on a path an agent printed in a pane (docs/58) ignores it: `activate_link` always goes through `open_file_at`, which is hard-wired to a tab. The doc comment says so deliberately — "Always a tab: File click behavior governs the FILES tree, not path activation" — but that was written when a tab *was* the click behavior.

The result is a setting that applies to half of the file clicks. Click a row in FILES: preview beside you. `Ctrl`+click the same path in the agent's output: a whole tab that hides the agent. The second case is the one that matters most — an agent that says "I wrote the plan to `docs/plan.md`" is giving you a place to click, and what you want is to see the file *next to* the conversation, not instead of it.

**What would you like to happen?**

`Ctrl`+click on a file path uses the same vocabulary as the FILES tree:

| Gesture | Today | Proposed |
|---|---|---|
| `Ctrl`+click on a path | tab (or the configured editor) | **File click behavior** — Preview by default; Tab keeps today's behavior, editor included |
| `Ctrl`+`Shift`+click | tab | a permanent split pane beside the focus, like `Shift`+click in FILES |
| right-click → Open File | tab | same as `Ctrl`+click |

`path:line` keeps jumping to the line in every target. URLs are unchanged.

Why not `Alt`+click for "open beside": links only light up while `Ctrl` is held (hover scanning is gated on it to stay off the hot path), so it would have to be `Ctrl`+`Alt`, and `Alt`+click is the classic window-manager drag gesture that terminals lose regularly. `Shift` already means "permanent pane" in the tree.

**Area**
files, input

**Platform**
all

**Alternatives**

An agent that knows about luvus can already run `luvus files open <path> --target preview`. This is for the far more common case where it just prints the path — no agent-side knowledge needed.

The change is small and lives in the #164 area: `activate_link` (`src/app/input.rs`) resolves `file_click_target()` (or `Pane` when the press carried `Shift`) and routes Preview/Pane through `open_file_view`, Tab through `open_file_at` as today, plus the line scroll on the preview. Docs: the "Opening links and files" table in the layout guide. I'll send a PR.

riclib:
Closes #185.

Since #164, **File click behavior** decides where a FILES row click lands — the reused preview pane by default, or a tab. A `Ctrl`+click on a path an agent printed in a pane (docs/58) ignored it and always opened a tab, which is the one case where a tab hurts most: the agent that printed the path disappears behind it.

## What changes

| Gesture | Before | After |
|---|---|---|
| `Ctrl`+click on a path | tab (or the configured editor) | **File click behavior** — Preview by default; Tab keeps the old behavior, editor included |
| `Ctrl`+`Shift`+click | tab | a permanent split pane beside the focus, like `Shift`+click in FILES |
| right-click → Open File | tab | same as `Ctrl`+click |

`path:42` preserves the requested line in Preview, permanent-pane and built-in read-only tab placements. A configured terminal editor opens the file at the top, because line-jump arguments differ between editors — `open_file_at` deliberately does not pass the line on. URLs are untouched.

## How

- `LinkPress` records whether `Shift` was down at the press (the press is the gesture; `Ctrl`+drag still hands over to the divider resize as before).
- `activate_link` resolves `file_click_target()` and delegates to a new `activate_link_in(target, place)`; the mouse-up arm passes `Pane` for the Shift variant. The context menu's Open File goes through `activate_link`, so it follows the setting too.
- `App::open_file_link` (files.rs) routes `Tab` to `open_file_at` unchanged — still the only path that can spawn an editor — and `Preview`/`Pane` to `open_file_view`, then scrolls whichever file view that left focused, checked against the path.
- Docs: the "Opening links and files" section of the layout guide now describes the placement and the Shift gesture.

## Tests

- `ctrl_click_on_a_file_path_previews_it_beside_the_pane` — default: no new tab, the focus is the workspace's preview view.
- `ctrl_click_on_a_file_path_opens_a_tab_when_that_is_the_click_behavior` — `layout.file_click = "tab"`: one new tab, not a preview (the previous test, now under its setting).
- `ctrl_shift_click_on_a_file_path_opens_a_permanent_pane_beside` — split, not a preview.
- `a_line_suffix_scrolls_the_viewer_to_that_line` passes unchanged under the new default, covering the line jump on the preview path.

`cargo fmt --all --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test --locked` (1113 passed) on Linux.

Known limitation, pre-existing: terminals that claim `Shift`+click for native selection never deliver the `Ctrl`+`Shift` variant (#170) — same as `Shift`+click in FILES today.

## Summary by CodeRabbit

- **New Features**
  - File links now follow the configured **File click behavior**, including preview panes, tabs, or the configured viewer/editor.
  - `Ctrl`+`Shift`+click opens a file in a permanent pane beside the current pane.
  - File links with a line number open at that line when supported; configured terminal editors open at the top.
  - Web links continue opening externally.

- **Documentation**
  - Updated layout guidance to explain file-link placement and keyboard shortcuts.

### DIFF
--- a/src/app/files.rs
+++ b/src/app/files.rs
@@ -271,12 +271,14 @@ impl App {
     }
 
     /// What a plain left click on a FILES row does, from `layout.file_click`
-    /// (docs/38). `Preview` is the default: one reused native read-only pane,
-    /// VS Code style. `Tab` is what a click did before this setting existed
-    /// and the *only* mode that consults `layout.file_open`, so it is also the
-    /// only mode that can launch an editor PTY. An unrecognized value — a config
-    /// touched by a newer Luvus — reads back as the default rather than leaving
-    /// a click doing nothing.
+    /// (docs/38) — and a `Ctrl`+click on a path printed in a pane (docs/58),
+    /// which is the same "open this file" with a different pointer. `Preview`
+    /// is the default: one reused native read-only pane, VS Code style. `Tab`
+    /// is what a click did before this setting existed and the *only* mode
+    /// that consults `layout.file_open`, so it is also the only mode that can
+    /// launch an editor PTY. An unrecognized value — a config touched by a
+    /// newer Luvus — reads back as the default rather than leaving a click
+    /// doing nothing.
     pub fn file_click_target(&self) -> OpenTarget {
         match self.config.layout.file_click.trim() {
             crate::config::FILE_CLICK_TAB => OpenTarget::Tab,
@@ -308,9 +310,10 @@ impl App {
 
     /// Open `path` in a new **tab** (docs/38), through the configured viewer:
     /// read-only or a terminal editor. This is the "open in tab" click
-    /// behavior, the fuzzy finder's action, and what `Ctrl`+clicking a path
-    /// printed in a pane always does (docs/58) — a *preview* click never comes
-    /// through here, which is why it can never reach an editor.
+    /// behavior, the fuzzy finder's action, and the tab half of `Ctrl`+clicking
+    /// a path printed in a pane (docs/58, [`open_file_link`](Self::open_file_link))
+    /// — a *preview* never comes through here, which is why it can never reach
+    /// an editor.
     ///
     /// `line` scrolls the built-in viewer to that line. It survives the async read
     /// because `FileView::apply` keeps `scroll` and clamps it to the file's length.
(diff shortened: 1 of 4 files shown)


RizRiyz:
@riclib One documentation correction is needed.

The CodeRabbit thread correctly found a contradiction, but its proposed direction does not match the implementation. In `src/app/files.rs`, `open_file_at()` intentionally does not pass `line` to a configured terminal editor, so terminal editors still open at the top.

Update `website/src/content/docs/docs/guides/layout.mdx` and the PR description to say:

> `path:42` preserves the requested line in Preview, permanent-pane, and built-in read-only tab placements. A configured terminal editor opens the file at the top because line-jump arguments differ between editors.

Also adjust the `open_file_link()` rustdoc so it does not imply every `Tab` target preserves the line.

The placement behavior itself looks correct and CI is green.

Separately, respond to the Apache 2.0 licensing confirmation requested here:
https://github.com/RizRiyz/luvus/issues/185#issuecomment-5465366435

riclib:
Fixed in 676df4d, in the direction you specified rather than the one the bot proposed.

Confirmed the behaviour first: with an editor configured, `open_file_at` calls `open_file_in_editor(path, &cmd)` and drops `line` on the floor. So the paragraph further down was the correct one, and taking CodeRabbit's suggestion would have made the docs describe a line jump that does not happen.

Three places updated:

- **`layout.mdx`** — the placement section now says the line is preserved in Preview, permanent-pane and built-in read-only tab placements, and that a configured terminal editor opens at the top because line-jump arguments differ between editors.
- **PR description** — same correction, replacing "lands on line 42 in every placement".
- **`open_file_link` rustdoc** — no longer opens with "jumping to `line`" as if it were unconditional. It now states that `line` is honoured in every placement ending in a native view, that an editor does not receive it, and therefore that a `Tab` target preserves the line only when `Open files with` is the built-in viewer.

One thing I did not do, so you can tell me if you would rather I did: the compiler-output paragraph lower in that section still explains the terminal-editor exception too, so the fact is now stated twice within about ten lines. Both are correct and they sit on different reader paths, so I left the second alone rather than editing text this PR had no business touching. Happy to trim it to a cross-reference if you prefer it said once.

`cargo fmt --all --check` and `cargo clippy --all-targets -- -D warnings` clean, 1113 passing.

The Apache 2.0 confirmation is posted on #185.

PR #220, opened 2026-08-31
fix(detect): parse padded ps process columns

### LINKED ISSUE #219: Agent detection misses commands when ps pads short PIDs
## Summary

On macOS, a pane can be classified as its shell and disappear from the Agents dock when its process IDs are shorter than the fixed-width `ps` columns.

## Reproduction

1. Run a recognized agent in a Luvus pane whose shell and agent PIDs have fewer than five digits.
2. Inspect the pane with `luvus pane processes <id> --json` and `luvus agent explain <id> --json`.

For a live pane running OpenCode, macOS reported:

```text
 5555 91833 /bin/zsh
 6647  5555 opencode2
```

Luvus exposed these executable identities instead:

```json
["91833", "5555"]
```

The pane was consequently classified as `zsh` with `command_fallback` authority, even though PID 6647 was the live `opencode2` child. Since `zsh` is not an agent, the pane was omitted from the Agents dock.

A pane with five-digit PIDs was detected correctly, making the behavior depend on PID width.

## Expected behavior

The process table parser should strip `ps` column padding, preserve the complete command field, and expose `zsh` plus `opencode2`. The pane should then be detected as OpenCode through `process_tree` authority.

## Cause

The macOS/BSD `ps` fallback parses PID and PPID with `split_whitespace`, but obtains the command with `line.splitn(3, char::is_whitespace)`. Leading or repeated alignment spaces count as delimiters and shift a numeric column into the command field.

## Environment

- Luvus 0.13.2 at `3d7e9a1`
- macOS arm64
- OpenCode executable: `opencode2`

rsaulo:
Preserve agent process identities when `ps` pads PID and PPID columns.

## What

- Parse PID and PPID after removing fixed-width alignment padding.
- Preserve the complete command field, including internal argument spacing.
- Add regression coverage for padded four-digit and unpadded five-digit process IDs.

## Why

On macOS/BSD, leading or repeated spaces in `ps -Ao pid=,ppid=,args=` could shift a numeric column into the command field. Recognized agents were then classified as their shell and omitted from the Agents dock.

Fixes #219.

## Verification

- [x] `cargo test ps_command_line_handles_padded_numeric_columns -- --nocapture`
- [x] `cargo test process_tree_finds_this_process_and_its_children -- --nocapture`
- [x] `cargo build --release --locked`
- [ ] `cargo test --locked` — 1227 passed and 4 unrelated tests failed; the same 4 failures reproduce on unmodified `origin/main` in this worktree (1226 passed).

## Notes

Validated on macOS arm64. Linux normally uses `/proc`; Windows uses its separate process snapshot implementation. The production server was not restarted during verification.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved process command parsing when numeric columns contain leading or padded whitespace.
  * Preserved spacing within command text for more accurate process information.

* **Tests**
  * Added coverage for process entries with padded numeric columns.

### DIFF
--- a/src/platform.rs
+++ b/src/platform.rs
@@ -552,6 +552,21 @@ fn proc_fs_table() -> Option<PsTable> {
 /// The process table from one `ps` invocation — the portable fallback and the
 /// path macOS/BSD always take. See [`ps_table`] for why Linux prefers `/proc`.
 #[cfg(unix)]
+fn parse_ps_command_line(line: &str) -> Option<(u32, u32, &str)> {
+    let line = line.trim_start();
+    let pid_end = line.find(char::is_whitespace)?;
+    let (pid, rest) = line.split_at(pid_end);
+    let rest = rest.trim_start();
+    let ppid_end = rest.find(char::is_whitespace)?;
+    let (ppid, command) = rest.split_at(ppid_end);
+    let command = command.trim_start();
+    if command.is_empty() {
+        return None;
+    }
+    Some((pid.parse().ok()?, ppid.parse().ok()?, command))
+}
+
+#[cfg(unix)]
 fn ps_command_table() -> Option<PsTable> {
     use std::collections::HashMap;
     let out = match std::process::Command::new("ps")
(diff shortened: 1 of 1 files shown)


RizRiyz:
Could you share before and after screenshot/video of this detection fix? appreciate the PR @rsaulo

rsaulo:
Sure, here is a before/after from the same macOS setup.

Before and after: the active OpenCode pane is missing before the fix and detected after it

Before: the active pane running `opencode2` was absent from the Agents dock. `pane processes` exposed the numeric columns as executables (`["91833", "5555"]`), so `agent explain` fell back to `zsh` via `command_fallback`.

After: I rebuilt this PR and restarted the server; the active pane now appears as `opencode` and its status updates normally. `pane processes` returns `["zsh", "opencode2"]`, while `agent explain` reports `identity.source: "process_tree"` and `status: "working"`.

The server restart renumbered the panes, so the before/after pane IDs differ.

PR #223, opened 2026-08-31
feat(modules): add telegram-notify example module (card t1)

sbkim:
A new example module that sends a Telegram message when one of your agents goes `blocked` or finishes (`done`), plus a right-click **Send a test message** action on any live agent.

## What

- `examples/modules/telegram-notify/`: `luvus-module.toml` (bot token as a secret, chat id, `notify_on` = blocked/done/both) + `notify.sh` (POSIX sh + curl, `--max-time 10`, fail-closed status gating, no retries)
- One plain `sendMessage` per status change; failures land in `luvus module log` only — the bot token never appears in argv or in any log line (the request URL is handed to curl via its stdin config)
- README with BotFather setup, an isolated-home live smoke test procedure (env-isolation prefix included so the smoke never touches a production server), and a troubleshooting section
- Index row in `examples/modules/README.md`

## Why

Agents occasionally need you when you are not at the desk (`blocked`) and you want to know when they finish. This gives a mobile notification path with zero dependencies beyond the shell.

## Verification

- [x] `sh -n notify.sh` on macOS `/bin/sh` and Alpine busybox `sh` (container)
- [x] ACs verified by direct invocation: gating, payload/fallback extraction, 401 on dummy token over real network, secret hygiene grep, argv/stdin evidence
- [x] Manual live smoke in an isolated dev home: real credentials, status-change and test-action messages delivered (operator-confirmed)
- [ ] `cargo test --locked` / clippy / fmt — no Rust changes in this PR; CI covers them

## Notes

- The live smoke surfaced one real-world robustness gap, fixed in this PR (32f4080): a pasted token with a stray space makes the assembled API URL malformed (`curl exit 3`) — settings are now trimmed, a mid-value space fails early with a clear message, and the curl exit-3 log line gained a hint.
- Module declares `platforms = ["macos", "linux"]`; no `src/` behavior changed.
- Optional follow-ups recorded in review: control-char escaping in `json_field` (F1), chat id/message visibility in curl argv up to the 10s timeout (F2).

🗿 MoAI

### DIFF
--- a/examples/modules/README.md
+++ b/examples/modules/README.md
@@ -1,6 +1,6 @@
 # Example modules
 
-Five complete, working luvus modules across three languages, each covering a
+Six complete, working luvus modules across three languages, each covering a
 different part of the extension surface. They are meant to be **copied and
 edited**, not installed as-is.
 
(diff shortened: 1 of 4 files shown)


RizRiyz:
@sbkim sorry we dont accept new module example PR, you can publish this as new repo and add 'luvus-module' in github topic for discovery. 

https://luvus.dev/docs/extend/writing-modules/#distribution--discovery

PR #234, opened 2026-09-01
feat(worktrees): list every checkout in Open Worktree

suixinio:
"Open Worktree" now lists every checkout of the repo — wherever it lives on disk and whichever tool created it — instead of only the folders under `~/.luvus/worktrees/`.

## What

- The workspace menu's **Open Worktree** opens a dedicated list modal fed by `git worktree list --porcelain`, replacing the filesystem folder picker rooted at `~/.luvus/worktrees/<repo>/`.
- Rows show the branch (short head when detached), the path, ⌂/⎇ markers for main vs. linked checkouts, and an "open" badge when the checkout is already a workspace.
- ⏎ opens the highlighted checkout as a workspace — or focuses the existing workspace instead of duplicating it (both paths canonicalized, so symlinked cwds still match). ↑/↓ or j/k move, esc closes; the ⏎/esc footer buttons stay clickable like the other modals.
- Bare entries and prunable leftovers (checkout folder already gone) are skipped; an empty result raises a toast instead of doing nothing silently.
- `Worktree` gains a `bare` flag parsed from the porcelain output; `is_main` keeps following git's ordering.
- New `no_worktrees_found` / `worktree_already_open` strings added to all nine locales.

## Why

The old "Open Worktree" only browsed the folder luvus itself creates worktrees in (falling back to the workspace cwd when that folder didn't exist), so any worktree added by hand (`git worktree add ../feature`), by another tool, or at any other path never showed — even though `git::local::worktrees()` already knew about it. The menu item now uses git as the source of truth.

## Verification

- [x] `cargo test --locked` (1289 passed)

New tests: porcelain `bare` parsing; a sibling-path worktree (the layout the old picker never showed) lists and opens; ⏎ focuses an already-open checkout instead of duplicating it; a full TestBackend render pass asserts the modal title, branch rows, and open badge.

## Notes

- Developed and tested on Linux; no manual TUI pass on macOS/Windows.
- No screenshot attached — the modal mirrors the existing folder-picker/prompt styling, and rendering is asserted by the TestBackend test.
- `worktrees_dir_for` still keys the *creation* folder by the main checkout's folder name; after this change that only affects where new worktrees are placed, not visibility. Happy to follow up separately.

## Summary by CodeRabbit

* **New Features**
  * Added an Open Worktree dialog listing available non-bare checkouts.
  * Open or focus a selected worktree using keyboard, mouse, and scrolling.
  * Added loading and Git error states during checkout discovery.
  * Added support for identifying bare repositories in listings.
  * Added localized loading and worktree status messages.
* **Bug Fixes**
  * Prevented dialog input from reaching the underlying pane.
  * Improved rendering of paths containing control characters.
* **Documentation**
  * Clarified that Open Worktree lists existing non-bare Git checkouts.
* **Tests**
  * Added coverage for asynchronous loading, path handling, selection, and bare repositories.


RizRiyz:
@suixinio The feature direction looks good, but these items need addressing before merge:

1. `open_worktree_list()` currently runs `git worktree list` and filesystem path checks synchronously on the app event loop. Move discovery to a bounded background worker and return the result through `AppEvent` so slow repositories or filesystems cannot freeze connected clients.

2. `src/git/local.rs` parses newline-delimited `--porcelain` output through `String::from_utf8_lossy()`. Use `git worktree list --porcelain -z` and preserve paths safely so unusual worktree paths, including paths containing newlines, remain valid. Add a focused regression test.

3. Update `website/src/content/docs/docs/guides/worktrees.mdx` to explain that **Open Worktree** lists Git checkouts from any location and focuses an existing workspace instead of opening a duplicate.

Also add a screenshot or short before/after video to the PR body showing the previous folder-picker behavior and the new worktree list modal.

suixinio:
@RizRiyz Thanks for the review — all four items are addressed in the three commits on top of the branch (ebeeb9b, 239f757, 1417fe2).

**1. Discovery off the app loop** — 239f757

The modal now opens immediately in a loading state and the scan (`git worktree list` + the per-checkout `is_dir` filter) runs on a worker thread, landing through a new `AppEvent::WorktreeListLoaded`. I followed the shape the named-session menu already uses: a `generation` counter bumped on every open *and* close, so a result for a modal that was dismissed or reopened in the meantime is dropped instead of resurrecting it. A git failure is shown inside the modal; an empty listing still closes it with the existing toast. ⏎ is a no-op while loading. The synchronous `is_repo` gate on the menu item is gone too — the worker reports "not a git repository" in the modal.

One deliberate deviation from the ask: I did not add a worker pool. Every other off-loop git job in the codebase is a plain `std::thread::spawn`, and the generation guard already makes a stray thread's result a no-op. If you want a shared bounded worker for git work, I think that's a separate change that should cover `worktree add`/`remove` as well (they are still synchronous and considerably slower than `list`) — happy to open an issue for it.

Tests: event round trip, stale-result discard on close and on reopen, the error row, and the loading row. The nine existing tests apply the scan inline through a small helper.

**2. `--porcelain -z`** — ebeeb9b

`worktrees()` now asks for `-z` first and falls back to the newline form when git rejects the flag. `-z` only exists since git 2.36, and Ubuntu 22.04 LTS still ships 2.34, so switching unconditionally would have broken the listing there. The parser takes the field iterator so both forms share it. Two regression tests: a NUL-terminated fixture with a newline in a path, and an end-to-end one that creates a real worktree at a path containing a newline and asserts it lists intact.

I left `String::from_utf8_lossy` alone in this PR. It lives in the shared `run()` helper, and `worktree_add`/`worktree_remove` also pass paths through `to_string_lossy`, so a proper fix is a byte-returning `run` plus `OsStr::from_bytes` on Unix across the whole module. That's a separate, self-contained change; I can follow up with it if you'd like.

**3. Docs** — 1417fe2

Added an "Open one" section to `guides/worktrees.mdx` explaining that **Open Worktree** lists every checkout `git worktree list` reports — wherever it lives and whichever tool created it — and focuses an already-open checkout instead of opening a duplicate. Also noted that `~/.luvus/worktrees/` is only where luvus *places* new worktrees. Expanded the one-line menu entry in `guides/layout.mdx` and linked it to the worktrees page.

**4. Screenshots**

image

RizRiyz:
Can you provide the video that I asked to show how this works? thanks @suixinio appreciate it. 

After that we can safely merge this PR

suixinio:
@RizRiyz

RizRiyz:
Thats perfect @suixinio 

Can you do a quick fix on the new git worktree modal to allow closing the modal by clicking anywhere outside the modal just like the open worktree UX? appreciate if you can do that and I will merge it after the fix. Thank you

suixinio:
@RizRiyz

PR #258, opened 2026-09-03
fix(detect): treat OMP π : title as working (incl. Windows brand mojibake)

Doeye1997:
## Summary
- #252 taught Luvus to read OMP OSC titles, but treated working as a braille spinner after `π`.
- Live OMP 18.1.x uses `π : label` while working, `π !` blocked, `π >` idle.
- Add an explicit `π :` Working title rule (priority 125) and keep the legacy spinner rule (124).
- On Windows ConPTY, live inventory shows the brand glyph as **U+87FA** (UTF-8 `E8 9F BA`) instead of Greek pi **U+03C0**, while ASCII state markers (`:`, `!`, `>`) stay intact. Match both brand codepoints so sidebar status classifies until the title encoding path is fixed.

## Test plan
- [x] `cargo test --locked detect::` — 47 passed, including `omp_title_states_work_without_the_optional_integration` with both `π` and `\u{87FA}` samples
- [ ] After install: `luvus server restart` (not only `server stop`) so the new binary loads
- [ ] Working OMP pane → `agent explain` shows Working / `manifest_rule`
- [ ] Idle OMP pane with `U+87FA > …` title → Idle via `manifest_rule`

## Validation
Unit coverage: `π :` / spinner / `!` / `>` plus Windows-mangled brand `U+87FA` variants. Pi isolation unchanged.

## Note
A deeper fix is still needed for OSC title decoding on Windows (non-ASCII title text such as Chinese is also garbled in `terminal_title`). This PR only unblocks OMP state classification.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OMP OSC-title state detection on Windows to recognize the ConPTY-mangled brand character alongside Greek pi.
  * Detection now correctly identifies blocked, working, idle, and legacy spinner title markers while preserving support for existing Greek pi and older spinner formats.
  * This improves status display reliability in Windows terminals using ConPTY, ensuring terminal indicators continue to reflect the correct operation state.

### DIFF
--- a/src/detect.rs
+++ b/src/detect.rs
@@ -431,38 +431,43 @@ fn builtin_rules() -> Vec<Rule> {
         // OMP publishes its state in the OSC title as `π <state> label`.
         // Current builds use `:` while working, `!` for attention, and `>` for
         // the user's turn. Older builds animated a braille spinner after `π `
-        // instead of `:`; keep that form so both contracts classify. Scope the
-        // rules to OMP: the generic spinner rule requires a line-leading
-        // spinner, and weakening it would create false positives for ordinary
-        // branded titles. The explicit idle title outranks retained screen
-        // activity but not a live confirmation panel.
+        // instead of `:`; keep that form so both contracts classify.
+        //
+        // On some Windows ConPTY paths the brand glyph arrives as U+87FA
+        // (UTF-8 E8 9F BA) instead of Greek pi U+03C0. Live inventory on this
+        // host shows that consistently while ASCII state markers stay intact,
+        // so match both brand codepoints until the title encoding path is fixed.
+        // Scope the rules to OMP: the generic spinner rule requires a
+        // line-leading spinner, and weakening it would create false positives
+        // for ordinary branded titles. The explicit idle title outranks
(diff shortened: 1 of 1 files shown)


RizRiyz:
Did you test this @Doeye1997 ? can you provide before and after video/screenshot of this bug?

Doeye1997:
Tested on Windows. Two screenshots from this machine (I’ll attach the JPGs here).

**Before** — `origin/main` at `66e7f0f` (#252 only: spinner / `π !` / `π >`, no `π :` working rule). Isolated session, OMP is mid-turn (tool calls). Sidebar: `idle omp`.
剪贴板图片 (73)


**After** — build with the `π :` Working title rule (prio 125). Same kind of live OMP turn. Sidebar: `working omp` on the busy pane; idle panes stay idle.
剪贴板图片 (74)


CLI `v0.13.4` is the same on both binaries; the difference is the detect rules, not the version string.

RizRiyz:
I am testing it on mac as well, now we just need to fix the CI so we can merge it @Doeye1997 

image

Issue #194, opened 2026-08-30
AGENTS: persist the All/Active sidebar filter


liao666brant:
**What problem would this solve?**

The AGENTS dock resets to **All** after every fresh start and snapshot restore. For users with many resumable sessions, historical rows can push live agents down the list, so **Active** has to be selected again after each restart.

In Luvus v0.13.2, `agents_active_only` is initialized to `false` in both app construction paths and is not persisted as a user preference:

- [fresh app initialization](https://github.com/RizRiyz/luvus/blob/v0.13.2/src/app/mod.rs#L2100-L2104)
- [snapshot restore initialization](https://github.com/RizRiyz/luvus/blob/v0.13.2/src/app/mod.rs#L2652-L2656)

This also conflicts with the [Working with Agents guide](https://github.com/RizRiyz/luvus/blob/v0.13.2/website/src/content/docs/docs/guides/agents.mdx#L40-L41), which says live agents only is the default.

**What would you like to happen?**

Persist the user's **All / Active** selection across restarts, or expose it as a configuration/Settings option. After a user selects **Active**, that preference should remain effective after restarting or restoring Luvus.

Keep **All** available for users who want to browse resumable history. This request does not require changing session discovery or stored sessions.

**Area**

agents, settings, website

**Platform**

all

**Alternatives**

The current workaround is to click **Active** after every restart. Patching both initialization sites and rebuilding works locally, but the change is lost on updates.

This cannot live as a module because the filter state belongs to the built-in AGENTS dock.

RizRiyz:
@liao666brant This is valid. The **All / Active** selection should remain as the user selected it after restarting Luvus or restoring a session.

We will persist this as a user preference in `config.json` and apply it to both fresh startup and snapshot restoration. Existing configurations without the preference will default to **All**, while future restarts will preserve the last selected value.

The fix should cover both the clickable toggle and the keyboard shortcut. We will also correct the Agents guide so the documented behavior matches the implementation.

This will not change agent detection, session discovery, or stored resumable sessions.

PR #261, opened 2026-09-03
feat(agents): scope agent list to one workspace

chuxubank:
`luvus agent list` can now be scoped to one workspace instead of always returning the whole session.

## What

- `agent.list` accepts an optional 0-based `workspace` (`node` is an accepted alias) or a stable `workspace_id`. Omitting both keeps the existing behavior and output shape exactly.
- Scope resolution reuses the shared workspace selector, so combining two selectors, or naming an unknown id, fails the same way as every other workspace method.
- An out-of-range index returns `not_found` rather than an empty list, and an unknown parameter returns `invalid_request`. A filter that silently widens back to the whole fleet is the failure mode that matters here: a script meaning to address one project would broadcast to every agent instead.
- CLI: `luvus agent list [--workspace <i> | --node <i> | --workspace-id <id>]`. Arguments are consumed explicitly, so a mistyped flag, a missing value, or a stray positional argument is refused before any socket connection.
- UHP parity: new `agentListParams` definition plus the `agent.list` `allOf` entry in `request.schema.json`, 3 valid and 4 invalid fixtures with updated manifest counts, and the matching rule in the dependency-free `examples/uhp/consumer.py` validator.
- Docs and bundled guidance updated: CLI reference, UHP methods, agent-messaging guide, both skill copies, and `agent-readme.md`. The help description is retranslated for all nine registered languages, none partial.

No response field was added or removed, and no existing method changed shape.

## Why

Every `agent.list` row already carries `workspace`, `workspace_name`, and `tab`, but the method itself walked all workspaces unconditionally and discarded whatever params it was handed. Anyone driving a multi-project session had to pull the entire fleet and filter client-side on every poll, and there was no server-side way to say "just this project".

No existing issue tracks this. I searched open and closed issues and PRs before starting; the nearest match, #194 / #197, is the sidebar All/Active state filter, which is a different axis.

## Verification

- [x] `cargo test --locked` — 1,334 passed, 0 failed, 1 ignored
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] `cargo fmt --all --check` — clean

Also run:

- `python3 examples/uhp/consumer.py` — validated 68 UHP fixtures and all JSON schema documents
- `python3 examples/uhp/terminal/consumer.py --fixtures` — validated 55 fixtures, all JSON schema documents, and snapshot replay
- `git diff --check` — clean

New focused tests: `agent_list_scopes_to_one_workspace_and_rejects_bad_filters` covers index, the `node` alias, `workspace_id`, an agent-free workspace, an out-of-range index, conflicting selectors, and an unknown parameter. `agent_list_accepts_one_workspace_scope` covers CLI parsing and every parse-time rejection.

The hand-rolled `consumer.py` validator and `request.schema.json` are two independent implementations, so passing the former does not prove the latter is correct. I cross-checked the schema with a real Draft 2020-12 validator (`jsonschema` 4.26) over 12 cases plus all published `agent.list` fixtures: zero disagreements with the reference validator.

Live run against a debug server in an isolated home (`LUVUS_HOME=~/.luvus-dev`, dedicated `--session agent-list-filter`, inherited socket and session selectors cleared, server stopped and temp dirs removed afterwards; no production session touched). Two workspaces with one reported agent each:

```
unscoped        -> pane 1 (claude, ws 0), pane 2 (codex, ws 1)
--workspace 0   -> pane 1
--workspace 1   -> pane 2
--node 1        -> pane 2
--workspace-id  -> the matching workspace only
--workspace 9   -> not_found: workspace 9 not found
--workspace-id nope -> not_found: workspace id nope not found
```

Parse-time rejections, all before connecting:

```
--workspace-nmae docs           -> unexpected agent list argument `--workspace-nmae`
--workspace                     -> --workspace needs a value
extra                           -> unexpected agent list argument `extra`
--workspace 1 --workspace-id b  -> --workspace and --workspace-id cannot be used together
--workspace docs                -> --workspace must be a 0-based number
```

## Notes

**Interaction with #257.** That PR replaces the same `("agent", _)` parser arm with `("agent", "" | "list") if rest.is_empty()`. Since `rest` is every argv token after the verb, including flags, `agent list --workspace 1` would be rejected as an unexpected argument if #257 landed as-is on top of this. The two changes are complementary rather than competing, so I deliberately made this arm strict in the same spirit and matched its `unexpected agent list argument ...` wording. Whichever merges first, reconciling them means keeping #257's unknown-verb arm alongside the flag-parsing list arm here; I did not touch the unknown-verb path, so that decision stays with #257.

Branched from `71f83d4` (#252), three commits behind `main` at the time of writing (#238, #251, #258). None of those touch any file in this diff, so it applies cleanly; happy to rebase if you would rather have it on current `main`.

Scoped to the workspace axis only. A `--tab` filter is the natural next step but is a separate user-facing change, so I left it out.

Verified on macOS only. The change is parameter validation and list filtering with no platform-specific code path.

## Summary by CodeRabbit

* **New Features**
  * Added workspace-scoped agent listings using workspace indexes, the node alias, or workspace IDs.
  * Unfiltered `agent list` continues to show agents across the session.
  * Added validation for supported filters, conflicting selectors, invalid indexes, and unknown parameters.

* **Documentation**
  * Updated CLI help, guides, reference documentation, examples, and translations.

* **Tests**
  * Added valid and invalid request coverage for workspace-scoped listings.

### DIFF
--- a/examples/uhp/consumer.py
+++ b/examples/uhp/consumer.py
@@ -14,6 +14,7 @@ AGENT = re.compile(r"^[a-z][a-z0-9_-]{0,31}$")
 REQUEST_ID = re.compile(r"^[A-Za-z0-9._:-]{1,128}$")
 SESSION_NAME = re.compile(r"^[A-Za-z0-9._-]{1,64}$")
 INTEGRATION_AGENT = re.compile(r"^[a-z0-9._-]{1,64}$")
+WORKSPACE_INDEX = re.compile(r"^[0-9]+$")
 EMPTY_HOST_METHODS = {
     "host.capabilities", "host.info", "host.doctor", "host.update.check",
     "session.list", "skill.status", "integration.status",
@@ -369,6 +370,26 @@ def valid_global_request(value, methods):
             and params["confirm"] is True
             and session_name(params["name"])
         )
+    if value["method"] == "agent.list":
+        params = value["params"]
+        if not set(params) <= {"workspace", "node", "workspace_id"}:
+            return False
+        for key in ("workspace", "node"):
+            if key not in params:
+                continue
+            scope = params[key]
+            if integer(scope):
+                if scope < 0:
+                    return False
+            elif not (isinstance(scope, str) and WORKSPACE_INDEX.fullmatch(scope)):
+                return False
+        if "workspace_id" in params and not bounded_string(
+            params["workspace_id"], 128, allow_empty=False
+        ):
+            return False
+        # One selector at a time: index, alias, or stable id.
+        if len(set(params) & {"workspace", "node", "workspace_id"}) > 1:
+            return False
     if value["method"] == "task.start":
         params = value["params"]
         if not set(params) <= {"id", "branch", "agent", "mode", "workspace_id"}:
--- a/plugins/luvus/skills/luvus/SKILL.md
+++ b/plugins/luvus/skills/luvus/SKILL.md
@@ -191,8 +191,12 @@ Resolve existing agents with:
 
 ```sh
 luvus agent list
+luvus agent list --workspace <i>
 luvus agent get <target>
 ```
+
+`agent list` covers the whole session. Scope it with `--workspace <i>` (0-based)
+or `--workspace-id <id>` when only one project's agents matter.
 
 Targets are a live name, pane id, or unique agent kind. If a kind is ambiguous,
 ask the user to choose or name the panes. Never guess.
--- a/protocol/uhp/v1/fixtures/invalid/requests.jsonl
+++ b/protocol/uhp/v1/fixtures/invalid/requests.jsonl
@@ -15,3 +15,7 @@
 {"id":"bad-session-extra","method":"session.start","params":{"name":"review","extra":true}}
 {"id":"bad-session-name","method":"session.start","params":{"name":"."}}
 {"id":"bad-integration-status","method":"integration.status","params":{"agent":"claude"}}
+{"id":"bad-agent-list-unknown","method":"agent.list","params":{"workspace_name":"docs"}}
+{"id":"bad-agent-list-both","method":"agent.list","params":{"workspace":"0","workspace_id":"workspace_example"}}
+{"id":"bad-agent-list-negative","method":"agent.list","params":{"workspace":-1}}
+{"id":"bad-agent-list-shape","method":"agent.list","params":{"workspace":"first"}}
--- a/protocol/uhp/v1/fixtures/manifest.json
+++ b/protocol/uhp/v1/fixtures/manifest.json
@@ -1,10 +1,10 @@
 {
   "protocol": { "name": "luvus-uhp", "major": 1, "minor": 0 },
   "files": [
-    { "path": "valid/requests.jsonl", "kind": "request", "expect": "valid", "count": 36 },
+    { "path": "valid/requests.jsonl", "kind": "request", "expect": "valid", "count": 39 },
     { "path": "valid/responses.jsonl", "kind": "response", "expect": "valid", "count": 5 },
     { "path": "valid/events.jsonl", "kind": "event", "expect": "valid", "count": 2 },
-    { "path": "invalid/requests.jsonl", "kind": "request", "expect": "invalid", "count": 17 },
+    { "path": "invalid/requests.jsonl", "kind": "request", "expect": "invalid", "count": 21 },
     { "path": "invalid/responses.jsonl", "kind": "response", "expect": "invalid", "count": 1 }
   ]
 }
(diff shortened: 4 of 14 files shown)


chuxubank:
Thanks — fixed in 9c7fe47. `--node` collapsed to `--workspace` for diagnostics, so `--node abc` reported the wrong flag name. Errors now quote the flag as typed:

```
--node abc       -> --node must be a 0-based number. usage: ...
--node           -> --node needs a value. usage: ...
--workspace abc  -> --workspace must be a 0-based number. usage: ...
```

`agent_list_accepts_one_workspace_scope` now pins this so the two flags cannot drift back to a shared diagnostic. Re-ran `cargo test --locked` (1,334 passed), `clippy --all-targets -- -D warnings`, and `cargo fmt --all --check`.

chuxubank:
Repositioning this after a reviewer pushed back with a fair question: if `agent.list` already returns `workspace` on every row, why not just `jq`?

For interactive use that objection lands, and I want to retract the implied convenience argument. `jq '.result.agents[] | select(.workspace=="1")'` is one line, and "saves a jq call" does not justify a parser, a schema definition, fixtures, and nine translations.

What I think does justify server-side scoping, narrowed down:

**1. It is the existing pattern for read-only methods here, not a new concept.** `mission.snapshot` already takes `workspace` / `workspace_id` and returns `not_found` for an out-of-range index (`dispatch.rs:3885`), and `git.status` resolves its cwd from a `workspace` param. I should have cited this in the description instead of arguing from first principles.

**2. An out-of-range scope is distinguishable from an empty one.** Client-side filtering returns `[]` for both "that workspace does not exist" and "that workspace has no agents". A script holding a stale index gets silence that reads like "nothing running over there". This PR returns `not_found`.

**3. No `jq` dependency.** Relevant for the Windows side of the matrix, where it is not a given and PowerShell wants a different idiom.

**One correction to something I got wrong.** I initially thought `workspace_id` filtering was impossible client-side. It is not: `workspace.list` returns both `workspace` and `workspace_id`, so a caller can join the two calls and then filter by index. That works. It is two round trips with the workspace set able to change in between — closing a workspace shifts every later index, since `workspaces` is a `Vec` and close goes through `Vec::remove` — so it is racy rather than impossible. Weaker claim than I made, and worth stating accurately.

**A smaller alternative, if you prefer it.** `agent.list` rows currently carry `workspace`, `workspace_name`, and `project`, but not `workspace_id`. Adding that one field would give client-side filtering a stable key with no new parameter, no CLI flag, no schema constraint, and no translation work — roughly three lines instead of this diff. It gives up the `not_found` signal and still needs `jq`, but it is a real option and I would rather name it than have it surface in review.

So: happy to keep this as-is, shrink it to just the `workspace_id` field, or close it. You have a much better sense than I do of how much query capability UHP should carry, and I would rather you pick than defend the larger version by default.

chuxubank:
Closing this in favor of a much smaller change. Two things I got wrong, and one I only saw after looking harder.

**The precedent I cited is a partial one, not an analogous one.** `mission.snapshot` takes `workspace` / `workspace_id` and returns `not_found`, which is what I quoted, but its motivation is limiting expensive per-workspace usage aggregation (`request_mission_usage_refresh_for`). `git.status` uses its `workspace` to choose the directory to run git in, which is targeting rather than filtering. `agent.list` is a plain in-memory walk, so scoping it server-side saves no work. Filtering a cheap read for the caller's convenience would be a new category here, and I presented it as an established one.

**The "no jq dependency" argument does not apply to UHP.** A UHP consumer is already parsing JSON in a real language, where this is a one-line filter. That reason only ever held for humans at the CLI, and I let it stand as though it justified the protocol surface too.

**The `not_found` argument is weaker than I claimed.** `workspace.get` already returns `not_found` for a bad index or id, so automation can validate first and filter client-side. What a single scoped call uniquely provides is atomicity: validate-and-filter in one request, with no window where a closing workspace shifts every later index. That is real but narrow.

**What decided it:** UHP has no general filtering or pagination concept for list methods. `pane.list`, `agent.list`, `task.list`, and `worktree.list` all return everything. Adding a one-off `workspace` parameter to one of them freezes a precedent into v1 that would sit awkwardly in front of any later uniform design for bounded or filtered lists, and every v1 parameter is permanent. That cost is not worth a narrow atomicity win.

The real gap underneath all of this is smaller and additive: `agent.list` rows carry `workspace` (a `Vec` index that shifts when an earlier workspace closes) and `workspace_name` (user-editable), but no stable key. So a consumer that wants to filter by a workspace that survives reordering has to join against `workspace.list` first.

Opening a follow-up that just adds `workspace_id` to each row. No new parameter, no schema constraint, no CLI flag, nothing that constrains a future list-filtering design.

Thanks to the reviewer who pushed back with "why not just `jq`?" — the honest answer turned out to be that most of what I built did not need to exist.

PR #264, opened 2026-09-03
feat(agents): publish the stable workspace id on agent list rows

chuxubank:
`agent.list` rows now carry `workspace_id`, so a consumer can identify an agent's workspace with a selector that survives reordering.

## What

- Each `agent.list` row gains `workspace_id`, the same stable id `workspace.list` and `workspace.get` already publish.
- Purely additive: no new parameter, no schema constraint, no CLI flag, and no change to any existing field or to `pane.agent_status_changed`.

## Why

A row identified its workspace two ways, and neither is stable:

- `workspace` is a 0-based position into a `Vec`. Closing an earlier workspace goes through `Vec::remove`, so every later index shifts down.
- `workspace_name` is the user-editable label.

So a consumer wanting to correlate agents with one project across those changes had to call `workspace.list`, find the index for its id, then filter `agent.list` by that index — two round trips, with the index able to change in between.

This is the reduced form of #261, which I closed. That PR added a server-side `workspace` / `workspace_id` filter parameter to `agent.list`. A reviewer asked why `jq` was not sufficient, and working through it honestly: the convenience argument did not hold, the precedent I cited (`mission.snapshot`) scopes to limit expensive per-workspace usage work rather than to filter a cheap in-memory read, and UHP has no general filtering or pagination concept for list methods — so a one-off parameter on one of them would have frozen an awkward precedent into v1. The only genuine gap left was the missing stable key, which is this.

## Verification

- [x] `cargo test --locked` — 1,333 passed, 0 failed, 1 ignored
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] `cargo fmt --all --check` — clean

Also run: `python3 examples/uhp/consumer.py` (61 fixtures) and `python3 examples/uhp/terminal/consumer.py --fixtures` (55 fixtures), both unchanged since no wire contract was constrained.

New test `agent_list_rows_carry_a_stable_workspace_id` asserts the id matches what `workspace.get` publishes for the same index, then reorders with `workspace.move` and asserts the index followed the move while the id did not.

Live run against a debug server in an isolated home (`LUVUS_HOME=~/.luvus-dev`, dedicated session, inherited socket and session selectors cleared, stopped and cleaned up afterwards). Two workspaces, one reported agent each, then closing workspace 0:

```
before close
   pane 1 | ws 0 | luvus        | workspace_7cf78f06…
   pane 2 | ws 1 | luvus-wsid-b | workspace_2bbae2e7…

after close
   pane 2 | ws 0 | luvus-wsid-b | workspace_2bbae2e7…
```

The survivor's `workspace` shifted from `1` to `0` while its `workspace_id` did not move, which is the case the field exists for.

## Notes

Branched from `71f83d4` (#252), a few commits behind `main`. Nothing in those commits touches `agent.list`; happy to rebase.

`pane.agent_status_changed` is deliberately untouched: its payload is pinned by `event-catalog.schema.json` with `additionalProperties: false`, so adding a field there is a separate contract change and not needed for this.

Verified on macOS only. The change is one field in a JSON response with no platform-specific path.

## Summary by CodeRabbit

* **New Features**
  * Agent listings now include a stable `workspace_id` for each agent, allowing reliable workspace identification even after reordering or renaming.

* **Documentation**
  * Updated agent messaging and UHP method documentation to explain workspace identifiers and recommend using `workspace_id` for persistent filtering and correlation.

### DIFF
--- a/src/app/dispatch.rs
+++ b/src/app/dispatch.rs
@@ -2449,6 +2449,11 @@ impl App {
                                 "state_source":s.state_source,
                                 "session": session,
                                 "workspace": wi.to_string(), "workspace_name": ws.name,
+                                // The one selector that survives reordering:
+                                // `workspace` is a positional index that shifts
+                                // when an earlier workspace closes, and
+                                // `workspace_name` is user-editable.
+                                "workspace_id": ws.id,
                                 "project": ws.name, "cwd": cwd,
                                 "branch": branch, "repo": repo, "worktree": is_worktree,
                                 "tab": (ti + 1).to_string(), "focused": id == focus,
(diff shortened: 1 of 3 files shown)


RizRiyz:
Approved! thanks @chuxubank

PR #266, opened 2026-09-03
feat(dock): optional tone and spans on module dock rows

screamyx:
Module dock rows can carry a Luvus Bar tone for their text, or a list of coloured spans, so a module can colour a row the same way it colours a bar segment.

## What

- `DockRow` gains `tone: Option<String>` and `spans: Vec<DockSpan>`. A row with neither renders exactly as before, in `subtext1`.
- `ui.dock.push` parses both from the row JSON:
  `{"text": "session [━━━━───] 72%", "tone": "success"}` or
  `{"text": "...", "spans": [{"text": "[", "tone": "muted"}, {"text": "━━━━", "tone": "success"}, {"text": "───]", "tone": "muted"}]}`.
- The sidebar draws spans left to right and truncates them as one line to the dock width. When the joined spans overflow, the last column is an ellipsis, the same as a plain text row, so a dropped tail (a `72%` after a bar that exactly fills the dock) is never silent.
- Tone names are the Luvus Bar ones (`normal`, `muted`, `accent`, `success`, `warning`, `error`) resolved through the existing `tone_color`, so rows follow every theme. Raw colours are still rejected. An unknown tone name falls back to the default row colour instead of failing the push.
- A span with no `text` is dropped at parse time, so a malformed `spans` list (`["a", "b"]`, `[{}]`) falls back to `text` instead of rendering a blank row. A row that sends `spans` and no `text` gets `text` from the spans joined in order, so `LUVUS_MODULE_ROW_TEXT` still names the row a click landed on.
- `BarTone::from_name` added with a unit test, plus a test that pins it to the serde spelling of every variant. `tone_color` is now `pub`. `ui::truncate` is split so its column-clipping half (`clip_columns`) can be reused without an ellipsis.
- Docs: the dock field table in *Writing a Module*, the `Row` schema in *UHP methods*, and the row shape in `skills/luvus-module/SKILL.md`.

## Why

A module dock could only colour its status dot. The row text was always `subtext1` and there was no field to colour it. A quota dock showing `[━━━━━━━───] 72%` per window had no way to show green, amber or red on the bar or the number, while the same module's Luvus Bar widget could. This gives docks the same vocabulary the bar already has, with the same theme-bound limits.

## Verification

- [x] `cargo test --locked`
- [x] Manual testing, on the final commit: linked a module that pushes the rows below into a release build under a separate `LUVUS_HOME` inside tmux, then read the cells back with `tmux capture-pane -e`. Every foreground in the QUOTA dock is one of the theme's `mint`, `amber`, `coral`, `overlay0` or `subtext1`, and each lands where the row asked for it: a `success` row is mint, a `warning` row's untoned spans are amber while its `muted` spans are overlay0, an `error` span is coral, a row with no tone and a row with the unknown tone `reddish` are both subtext1. The overflow row ends in `…` in the cut span's colour with the `72%` tail dropped, and nothing crosses the border. Pre-existing docks render unchanged.

Tests added:

- `dock_push_preserves_row_tone_and_spans` (dispatch): tone and spans survive the push, a span without a tone stays `None`, a row with neither keeps the old shape, an unknown tone is stored and left to the renderer.
- `row_tone_colours_the_text_and_untoned_rows_keep_the_default` (sidebar, rendered): success is the theme's mint, no tone is `subtext1`, an unknown tone is `subtext1`.
- `spans_take_their_own_tone_or_inherit_the_rows` (sidebar, rendered): a span with a tone uses it, a span without one uses the row's, and with no row tone the default.
- `spans_truncate_as_one_line_with_a_trailing_ellipsis` (sidebar, rendered): an overflowing tail is dropped with an ellipsis in the last column, an exact fit keeps its tail with no ellipsis, a single over-wide span is clipped, nothing is drawn past the dock.
- `a_wide_glyph_at_the_edge_ends_the_line_instead_of_being_skipped` (sidebar, rendered): a two-column glyph that does not fit ends the line, later spans do not slide into its place, the ellipsis takes a drawn span's colour.
- `dock_push_drops_empty_spans_and_backfills_text_from_spans` (dispatch): bare strings and empty objects are not spans, a spans-only row gets its text from the spans, an explicit text is never overwritten.
- `tone_from_name_matches_the_serde_spelling_of_every_variant` (bar): exhaustive, so a new tone cannot work in the bar while silently falling back in docks.

## Screenshots

Text capture of the dock from that run (quattro-rally theme):

```
  QUOTA
  5h  [━━━━━━━━──] 78%      success row, all mint
  7d  [━━━━──────] 41%      warning row: 7d, bar, 41% amber; brackets and ─── muted
  mo  [━━━━━━━━━━] 100%     no row tone: mo subtext1; bar and 100% coral; brackets muted
  plain untoned row         subtext1
  typo tone row             tone "reddish": subtext1
  ━━━━━━━━━━━━━━━━━━━━━━…   overflow: mint bar, mint ellipsis, " 72%" dropped
```

## Notes

- No new UHP method or schema. The two fields are optional keys on an existing `ui.dock.push` row object.
- `text` stays the plain string a click hands to the action as `LUVUS_MODULE_ROW_TEXT`; spans are presentation only.
- `tone: "normal"` is the bar's text colour (`subtext0`), one step dimmer than an untoned row (`subtext1`). Documented in both tables.
- Related: the bar `progress` segment cannot be pushed over JSON because its numeric `value` collides with the flattened click payload `value: Option<String>` (`"expected a string"` for an integer, `"missing field value"` for a string). Not touched here, tracked in #306 / #307.
- Implementation was tool-assisted. Everything under Verification was run by me.

## Summary by CodeRabbit

- **New Features**
  - Sidebar dock rows support configurable tones and independently colored text spans.
  - Rows can derive action text by joining valid spans when explicit text is omitted.
  - Supports standard tone names, including normal, muted, accent, success, warning, and error.

- **Bug Fixes**
  - Malformed or empty spans are safely ignored.
  - Improved clipping prevents later text from shifting into unavailable space, including at wide-glyph boundaries.
  - Unknown tone names use the default color.

- **Documentation**
  - Updated module and UHP references with tone, span, and empty-row behavior.


RizRiyz:
@screamyx The implementation looks good. Tone resolution follows the active theme, existing rows remain unchanged, span truncation is width-aware, and I found no security or dependency concerns.

Before merging, could you complete the public contract and test coverage:

- Update `website/src/content/docs/docs/uhp/methods.mdx` so the `ui.dock.push` row schema includes `tone` and `spans`
- Update `skills/luvus-module/SKILL.md`, which still describes rows as `{text, action?, value?}`
- Add a dispatch test confirming `ui.dock.push` preserves the row tone and spans
- Add coverage for span tone inheritance, unknown-tone fallback, and width truncation

After those additions, this should be ready to merge.

screamyx:
Done, all four:

- `methods.mdx`: the `Row` schema now lists `tone` and `spans`, with a paragraph on inheritance, the unknown-tone fallback, and why `text` still travels with `spans`.
- `skills/luvus-module/SKILL.md`: row shape updated to `{text, dot?, tone?, spans?, action?, value?, menu?}`.
- Dispatch test `dock_push_preserves_row_tone_and_spans`.
- Three rendered sidebar tests for span tone inheritance, unknown-tone fallback, and width truncation.

One behaviour change found while writing the truncation test: when the first span exactly filled the dock, later spans were dropped with no ellipsis. Spans now truncate as one line and reserve the last column for the ellipsis, matching plain text rows. The test pins that.

Rebased on current main.

screamyx:
Pushed a second follow-up that closes both CodeRabbit comments and a few things a local diff review turned up:

- `skills/luvus-module/SKILL.md`: `normal` added to the tone list, span tones use the same names, and an absent tone is stated to differ from `normal`.
- Span truncation now stops at the first span that does not fit whole. A two-column glyph at the edge used to be skipped while the spans after it slid left into its place. New rendered test pins it.
- A span with no `text` is dropped at parse time, so `"spans": ["a", "b"]` or `[{}]` falls back to `text` instead of rendering a blank row.
- A row that sends `spans` and no `text` now gets `text` from the joined spans, so `LUVUS_MODULE_ROW_TEXT` is never empty for a clickable row. Docs updated in both places.
- Test hygiene: the sidebar helper now returns its `TestEnv` guard so `$LUVUS_HOME` stays isolated for the whole test body, and `BarTone::from_name` has an exhaustive test against the serde spelling.

screamyx:
Fixed the contradiction: the methods reference now says a row needs `text`, or `spans` with at least one non-empty span, marks `text` optional in the schema block, and states what a row with neither does (renders empty, hands its action an empty string). The module guide's field table carries the same note.

RizRiyz:
Approved @screamyx

PR #268, opened 2026-09-04
feat(agents): scope the dock to the active workspace

### LINKED ISSUE #262: AGENTS: scope the sidebar dock to one workspace
**What problem would this solve?**

In a session that holds several projects, the AGENTS dock lists every agent from every workspace in one flat list, and the only way to tell them apart is by reading the workspace name on each row's second line.

I keep one session with a handful of checkouts open — an app, its API, a docs site — each with one or two agents running. When I am working in the API and an agent goes blocked, I have to scan the whole list and read workspace names to find the rows that belong to the project I am actually in. Once the dock scrolls, the agents I care about are not even on screen together.

The existing All/Active toggle does not help here: it filters by lifecycle (live agents vs. resumable history), not by project. `pinned_agents` is the closest thing today, but pinning is a manual per-pane action that says "these specific panes", not "this project", and it has to be redone as agents come and go.

**What would you like to happen?**

A way to narrow the AGENTS dock to a single workspace, so a multi-project session can show just the agents for the project I am working in.

I deliberately have no strong opinion on the interaction, and I would rather you pick the shape. Some options, roughly least to most ambitious:

1. The dock follows the focused workspace, with an explicit way to go back to all of them.
2. A second header control next to All/Active, e.g. All / This workspace.
3. Selecting a workspace in the WORKSPACES dock scopes the AGENTS dock. Reads the most natural to me, but it couples two docks' state, so it is the one most likely to have consequences you can see and I cannot.
4. Keep everything visible but group rows under workspace subheaders instead of filtering.

One follow-up either way: should the choice persist across a server restart, the way #197 made the All/Active toggle persist, or reset each time?

**Area**

agents, workspaces

**Platform**

all

**Alternatives**

A module is a real option now, not a hypothetical: #261 (open) adds `luvus agent list --workspace <i> | --workspace-id <id>`, so a module can poll that and push its own scoped dock with `ui dock push` the way `examples/modules/branch-dock` does. That gets the behavior without touching core.

I still think the built-in dock is the right home. This is a "which of these do I care about right now" problem in the dock people already have open, and solving it by asking users to install a module that duplicates AGENTS — losing pin, All/Active, and the existing click and right-click behavior — seems like the wrong trade. Reasonable to disagree, and if you would rather this stay out of core I am happy to write the module instead.

Workarounds today: read the workspace name on each row, pin the agents I am using, or drop to the CLI. After #261 the CLI can answer this precisely and the TUI cannot, which is the gap I am pointing at.

Not a duplicate of #194/#197 (persisting the All/Active lifecycle filter), #35 (pin agents and workspaces to the top, which reorders rather than narrows), or #237 (dock heights). The AGENTS dock builds its own list by walking `app.workspaces` in `draw_agents_dock`, independently of `agent.list`, so #261 does not change any UI behavior on its own.

Happy to implement whichever direction you prefer — I would just rather agree on the interaction first than send a PR that guesses at it.

chuxubank:
The built-in AGENTS dock can now show only the active workspace without changing its global attention model.

Closes #262.

## What

- Adds a persisted `agents_this_workspace` preference, defaulting to All workspaces.
- Keeps workspace scope independent from lifecycle: **All / Active** still controls resumable history, while **Here** controls which workspace contributes rows.
- Draws **Here** on a second header row only when more than one workspace is open. The default sidebar is 26 columns; after the title and right-aligned All/Active control only four columns remain, which does not fit ` Here ` and cannot fit several translations. The second-row option from #262 keeps the axes visually separate and works in every locale.
- Scopes both live agents and resumable sessions. Session ownership uses the longest matching open workspace root, so a nested workspace owns its own history instead of leaking into its parent.
- Keeps attention global. Detection, sounds, compact counts, and `focus_next_attention` are unchanged. If scoped rows hide blocked agents, one fixed overflow line reports the count; clicking it jumps directly to a hidden blocked pane and leaves the scope selected.
- Hides the scope chip with one workspace, while preserving and applying the saved choice to resumable history.
- Clears the new hit geometry when sidebars or workspaces disappear, and preserves the active client's geometry across passive client projections.
- Adds complete strings for all nine registered languages and updates the layout/configuration documentation.

## Why

A multi-project session currently mixes every live agent and resumable session into one flat dock. Workspace labels help identify rows, but do not help once the project's rows scroll off screen. Pinning expresses interest in individual panes rather than one project and has to be repeated as agents change.

The dock is only a view, so this intentionally does not change what Luvus considers attention-worthy. A blocked agent in another workspace still affects sounds and global counts and remains one click away.

## Verification

- [x] `cargo test --locked` — 1,361 passed, 0 failed, 1 ignored after rebasing onto current `upstream/main`
- [x] `cargo build --release --locked`
- [x] `python3 examples/uhp/consumer.py` — 61 fixtures
- [x] `python3 examples/uhp/terminal/consumer.py --fixtures` — 55 fixtures

Focused coverage:

- default All-workspaces behavior and chip hidden with one workspace;
- the independent scope/lifecycle axes at the default 26-column sidebar width;
- persisted click selection and config patch/reload;
- live-agent filtering;
- resumable-session filtering with nested workspace roots;
- blocked-overflow count and direct jump to a hidden pane without widening;
- stale hit-geometry cleanup;
- complete `{n}`-aware text in all registered languages;
- fresh and restored App construction.

## Notes

No CLI or UHP method changes. No keybinding was added; #262 asked for the visible chip, and adding a command would unnecessarily enlarge the public control surface.

No changes to attention navigation, sounds, or compact status counts; their files have zero diff in this branch.

## Summary by CodeRabbit

- **New Features**
  - Added an independent workspace filter for the AGENTS dock, toggled with `A` or from an agent/session context menu.
  - The filter shows live agents and resumable sessions in the active workspace while preserving global attention indicators.
  - Added an overflow indicator for agents in other workspaces, with click-to-jump navigation without clearing the filter.
  - Workspace and All/Active filter choices persist across restarts and are localized across supported languages.

- **Documentation**
  - Updated layout, configuration, and keybinding references with the new workspace-filter behavior.


chuxubank:
Addressed both CodeRabbit findings in `714a4fd`:

- scope/overflow hit geometry now resets above the tiny-window early return; `renders_across_phone_viewports` seeds stale targets and proves the 20×4 guard clears them;
- configuration docs now separate chip visibility from persisted behavior and explicitly state that one-workspace resumable history remains scoped by the remaining root.

Re-ran focused tests, strict Clippy, all 1,361 locked tests, and the locked release build.

chuxubank:
One interaction question before this merges: should workspace scope also get a prefix key?

My suggested pairing is:

- `Ctrl+Space a` — existing All / Active lifecycle toggle
- `Ctrl+Space A` — All workspaces / Here scope toggle

Uppercase `A` is currently unbound in the default keymap, and the lowercase/uppercase pair makes the two independent AGENTS axes discoverable without inventing another mnemonic. I left it out of the implementation for now because #262 explicitly described the visible chip and a new command/keybinding expands the configurable command surface, localization, docs, and tests.

Happy to add `ToggleAgentScope` + default `A` in this PR if you want keyboard parity, or keep scope mouse-only for v1.

RizRiyz:
Yeah we should have prefix for that, currently only` prefix+N` tied to create new workpace, other than that using `Shift+Number` are used to jump between workspace.

chuxubank:
Added the prefix key in `b7598e5` as discussed:

- prefix `a` remains All / Active;
- prefix `A` now toggles All workspaces / Here;
- new stable command id: `toggle_agent_scope`, exposed in Settings → Keys;
- labels translated for all nine registered languages;
- keybinding and layout docs updated;
- tests pin both default bindings and both persisted scope states.

Re-ran strict Clippy, all 1,362 locked tests, and the locked release build.

chuxubank:
One edge case surfaced during local preview: **Here still has a real effect with one open workspace** because the unscoped AGENTS dock may include resumable sessions whose cwd belongs to other projects. Live panes are identical either way, but history is not.

The current implementation follows your original “hide the extra chip when there is only one workspace” direction while preserving the saved scope. With one workspace:

- `agents_this_workspace=true` still filters resumable history to that workspace root;
- the Here chip is hidden;
- prefix `A` can still toggle it, but a mouse-only user cannot see or change the effective scope.

Which behavior do you want?

1. **Keep current behavior:** hide the chip; prefix `A` is the only control in the one-workspace case.
2. **Always show Here:** simplest and fully visible, because history makes the axis meaningful even with one workspace.
3. **Show Here only when it can change rows:** one workspace plus any resumable session outside its root. More precise, but the header can appear/disappear as discovery refreshes.
4. **Suspend scope with one workspace:** keep the chip hidden and show all history; preserve the preference for when a second workspace opens.

After seeing it live, I lean toward **2**. It avoids invisible effective state and keeps mouse/keyboard parity; the extra row costs space, but the filter is genuinely meaningful for history.


https://github.com/user-attachments/assets/92d06963-bd3f-4961-85fa-9a36174be0fe

RizRiyz:
@chuxubank I dont agree we add extra button in the UI like that since we already have All/Active button in the agents, lets stick to using prefix+A and dont dont forget to add that in help modal and allow the key to be edited in Keys settings.

Btw what softrware you used to run that luvus and screen recorder app you used? I'd like to use that for making luvus video, it looks beautiful.

chuxubank:
Got it — I’ll remove the clickable Here row and keep workspace scope on prefix `A`; it is already generated into the `?` help modal from `Cmd::ALL` and exposed as the editable `toggle_agent_scope` row in Settings → Keys.

I do think the effective scope needs passive visual feedback, otherwise `A` can hide rows without leaving any indication why. Would you be okay with a **non-clickable, no-extra-row marker** in the existing header only while scope is active — for example `AGENTS · HERE` (localized), while unscoped remains plain `AGENTS`? It would add no button, no hit target, and no persistent extra height.

For the video: the terminal is **WezTerm** + **Starship** and the screen recorder is **Recordly**.

RizRiyz:
@chuxubank I think for visual feedback we can add "Show Workspace Only/Show All Workspace" in the right click inside agents dock right click along with this, so that active can be active for workspace only or global
image

Thanks for apps info.

chuxubank:
Implemented your row-menu suggestion in `fc2cd34`:

- removed the extra clickable Here row entirely;
- right-clicking any live agent or resumable-session row now includes a dynamic action:
  - `Show Workspace Only` while global;
  - `Show All Workspaces` while scoped;
- the AGENTS title itself does not get a context menu;
- prefix `A` remains available in the `?` help modal and as an editable Settings → Keys command;
- both menu states are localized in all nine registered languages;
- the choice still persists and remains independent of All / Active.

Verified with focused row-menu/scope/keymap/i18n tests, strict Clippy, all 1,362 locked tests, and the locked release build.

chuxubank:
One CI job needs a rerun. `fmt · clippy · test` reached the full locked suite, then the unrelated platform test `unix_stoppable_pid_accepts_a_luvus_executable_with_arguments` failed while spawning its freshly written temp executable with Linux `ETXTBSY` (`Text file busy`). The run had 1,376 passes before that single failure; fmt/Clippy completed, all other checks are green, and the same full locked suite passed locally (1,362 tests on this checkout).

I tried `gh run rerun 33849144754 --failed`, but GitHub requires repository admin rights. Could you rerun the failed job? No product-code change appears warranted for this PR.

RizRiyz:
Approved!

PR #270, opened 2026-09-04
feat(session): add stop session via right-click context menu

aancw:
Right-click a running non-current session row to show Stop, stopping only that session via `session::stop_session(Some(name))` without affecting the active TUI.

## What

- Right-click on a session row in the Sessions switcher opens a context menu anchored at the click position; menu shows `Stop` only for running, non-current sessions (stopped sessions and current session show no menu).
- `Stop` runs `session::stop_session(Some(name))` off the app loop via `AppEvent::NamedSessionStopped`; targets isolated `api` + `client` sockets for that session name, so stopping session B does not stop or detach the active session.
- Right-clicking another row while the menu is open re-anchors the menu to the new row; right-clicking empty area, the `+ New Session` row, or the menu itself dismisses/keeps it consistently (matches Agents `Resume`/`Close` behavior).
- Menu drawn via `render_popup` (`PopupId::Session`) on top of the Sessions popup; closes on click outside, `Esc`, or Sessions popup close.
- i18n: `menu_stop_session` added for 9 locales (`Stop` / `Detener` / `Parar` / `Arrêter` / `Beenden` / `Hentikan` / `停止` / `停止` / `중지`).

## Why

Sessions switcher previously only had `+ New Session` with no way to stop a session from the UI. Users had to run `luvus session stop <name>` from a shell, which could be confused with the current session and lacks discoverability. This adds the same isolated stop logic to the UI with a familiar right-click pattern (used by Agents and other context menus).

## Verification

- [x] `cargo check` (0 warnings)
- [x] `cargo test` — 1377 passed; 1 pre-existing failure `app::tests::resize_yields_to_pane_title_and_zoom_but_still_grabs_the_seam` (also fails on `main`)
- [ ] Manual testing: open Sessions switcher with 2+ named sessions, right-click a running non-current row → `Stop` appears at click; click `Stop` → toast `stopped <name>`, row removed; right-click `item2` then `item3` → menu follows; right-click current/stopped → no menu

## Screenshots

luvus-context

## Notes

- `delete session` is not included; planned separately.
- Branch: `feat/session-stop-context-menu`

## Summary by CodeRabbit

- **New Features**
  - Added a context menu for running, non-current named sessions.
  - Added an option to stop a selected named session.
  - Added mouse and keyboard controls for opening, selecting, re-anchoring, and closing the menu.
  - Added localized “Stop Session” text across supported languages.

- **Bug Fixes**
  - Session lists now refresh after a session is stopped.
  - Stop results and errors are surfaced appropriately, with stale responses safely ignored.
  - Stop menus no longer appear for stopped or current sessions, or while session preparation is pending.
  - Menu input no longer affects session rows or scrolling behind the open menu.


RizRiyz:
@aancw The two original CodeRabbit findings are still valid and unresolved. The later “No actionable comments” result only reviewed the Clippy follow-up commit and skipped the previously reviewed code.

Please address both before merge:

1. Prevent the Stop menu from opening while `NamedSessionMenu::preparing` is true. Apply this guard to both right-click paths so an attachment in progress cannot be interrupted.

2. Clear `self.session_menu` when the selected row is current or stopped. This prevents the previous session’s Stop action and hitbox from remaining active.

Please add focused regression tests for the preparing-state guard and stale-menu dismissal, then complete the isolated two-session manual test described in the PR.

aancw:
> @aancw The two original CodeRabbit findings are still valid and unresolved. The later “No actionable comments” result only reviewed the Clippy follow-up commit and skipped the previously reviewed code.
> 
> Please address both before merge:
> 
>     1. Prevent the Stop menu from opening while `NamedSessionMenu::preparing` is true. Apply this guard to both right-click paths so an attachment in progress cannot be interrupted.
> 
>     2. Clear `self.session_menu` when the selected row is current or stopped. This prevents the previous session’s Stop action and hitbox from remaining active.
> 
> 
> Please add focused regression tests for the preparing-state guard and stale-menu dismissal, then complete the isolated two-session manual test described in the PR.

Please check, thanks

PR #275, opened 2026-09-05
fix(input): honor negotiated CSI-u for control keys

chuxubank:
Honor pane-negotiated Kitty keyboard disambiguation by encoding every Ctrl+character chord with CSI-u.

## What

- Preserve legacy control bytes, such as ETX for `Ctrl+C`, until the nested application enables `DISAMBIGUATE_ESCAPE_CODES`.
- Once negotiated, encode every Ctrl+character chord consistently with CSI-u, including `Ctrl+C`, `Ctrl+;`, and `Ctrl+'`.
- Keep the existing legacy `Ctrl+/`/`Ctrl+7` normalization while preserving distinct CSI-u identity where the host can report it.
- Add regression coverage for punctuation, digits, Unicode, legacy compatibility, and full negotiated Ctrl encoding.

## Why

Luvus previously dropped Ctrl+character chords without a traditional control-byte representation. A fallback-only fix handled those missing chords but continued mixing legacy control bytes with CSI-u after the nested application had explicitly negotiated Kitty keyboard disambiguation. Full gated encoding follows the negotiated protocol consistently while leaving ordinary shells unchanged.

## Verification

- [x] `cargo test --locked`
- [ ] Manual testing, if applicable

## Notes

The behavior remains pane-scoped and gated by the nested application's Kitty keyboard mode. Applications that do not request CSI-u retain the existing legacy input bytes and receive no new escape sequences.

## Summary by CodeRabbit

- **Bug Fixes**
  - Ctrl+character shortcuts now use consistent CSI-u encoding when enhanced keyboard reporting is negotiated, including combinations that previously used legacy control-byte encodings.
  - Shifted letters, including applicable Unicode characters, are normalized consistently so control and shifted-letter shortcuts remain distinct in nested terminal applications.

- **Tests**
  - Updated coverage for negotiated CSI-u encoding and distinct shortcut delivery through nested terminal applications.

### DIFF
--- a/src/app/input.rs
+++ b/src/app/input.rs
@@ -4054,8 +4054,11 @@ fn encode_key(
                         // slash binding for the nested CSI-u client.
                         '/' | '7' => '/',
                         // Kitty reports the unshifted codepoint and carries
-                        // Shift in the modifier parameter.
-                        'a'..='z' | 'A'..='Z' if shift => c.to_ascii_lowercase(),
+                        // Shift in the modifier parameter. Normalize Unicode
+                        // letters too when their lowercase form is one scalar.
+                        character if shift && character.is_alphabetic() => {
+                            single_lowercase_codepoint(character)
+                        }
                         _ => c,
                     };
                     return Some(csi_u_char(codepoint, key.modifiers));
@@ -4136,6 +4139,17 @@ fn encode_key(
         _ => return None,
     };
     Some(bytes)
+}
+
+/// Lowercase a key identity only when Unicode maps it to exactly one scalar.
+fn single_lowercase_codepoint(character: char) -> char {
+    let mut lowercase = character.to_lowercase();
+    let first = lowercase.next().unwrap_or(character);
+    if lowercase.next().is_none() {
+        first
+    } else {
+        character
+    }
 }
 
 fn legacy_control_byte(character: char) -> Option<u8> {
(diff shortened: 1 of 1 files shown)


RizRiyz:
@chuxubank Before merging, please add coverage for `Ctrl+Shift` with a non-ASCII letter.

The negotiated CSI-u path currently normalizes shifted ASCII letters with `to_ascii_lowercase()`, but leaves shifted Unicode letters unchanged. Please verify that an input such as `Ctrl+Shift+É` emits the unshifted `é` codepoint with the Ctrl+Shift modifier, and that the non-negotiated path remains unchanged.

This will make the new “every Ctrl character” behavior reliable across international keyboard layouts.

chuxubank:
@RizRiyz Addressed in e743c01. The negotiated path now lowercases shifted Unicode letters when the lowercase mapping is exactly one codepoint, so Ctrl+Shift+É emits CSI 233;6u. The non-negotiated path remains unchanged and produces no bytes for that chord. I also kept multi-codepoint lowercase mappings unchanged to avoid inventing an invalid single-key identity. Verified with cargo test --locked, cargo clippy --all-targets -- -D warnings, and cargo fmt --all --check.

RizRiyz:
Thanks @chuxubank 

Approved!

PR #307, opened 2026-09-07
fix(bar): accept numeric progress value alongside click payloads

### LINKED ISSUE #306: Luvus Bar progress segment cannot be published: numeric value rejected as "expected a string"
**What happened?**
The documented Luvus Bar `progress` segment cannot be published. The docs ("Luvus Bar" in the UHP methods reference, and the module guide) show:

```json
{"type":"progress","value":3,"total":7,"width":8}
```

but pushing it is rejected:

```text
$ luvus bar push --id status --content '[{"type":"progress","value":23,"total":100,"width":6}]'
invalid_request: invalid content: invalid type: integer `23`, expected a string
```

Sending `value` as a string fails the other way (`string "100", expected u64` for `total`), and any other numerator field name yields `missing field 'value'`. So there is no shape that publishes a progress bar. Expected: the documented shape validates (`0 <= value <= total`, `total > 0`, `width` 3..=24, default 8) and renders.

**Cause**
`BarSegment` in `src/bar/mod.rs` flattens the `type`-tagged `BarSegmentKind` enum and also declares its own `value: Option<String>` (the click payload used by text/symbol/state/badge). Serde drains the outer struct's fields from the buffered map first, so the numeric `value` for `Progress { value: u64, .. }` is claimed as a string before the enum sees it. Serializing a progress segment also emits `value` twice.

**Steps to reproduce**
1. Link any module that declares a `[[bars]]` widget, e.g. `luvus module link /path/to/luvus/examples/modules/ci-bar`.
2. `luvus bar push --id status --content '[{"type":"progress","value":23,"total":100,"width":6}]'`
3. Observe `invalid_request: invalid content: invalid type: integer `23`, expected a string`.

**Area**
bars, modules

**Environment**
- OS + version: Ubuntu 26.04 (Linux 7.0)
- Luvus version: 0.13.4 (installed release) and `main` @ c36d792 (reproduced with a debug build and an isolated `LUVUS_HOME`)

knowsuchagency:
The documented Luvus Bar `progress` segment (`{"type":"progress","value":3,"total":7,"width":8}`) now publishes instead of failing with `invalid type: integer, expected a string`.

## What

- `BarSegmentKind::Progress` is now the sole owner of the `value` key, so numeric `value`/`total`/`width` deserialize, validate, and render as documented. Validation and error strings are unchanged (`0 <= value <= total`, `total > 0`, `width` 3..=24, default 8).
- The string click payload `value` (used with `action`) moved from the outer `BarSegment` onto the variants that can carry one: `text`, `symbol`, `state`, `badge`, `spacer`, `separator`. Their JSON shape is unchanged, e.g. `{"type":"badge","text":"2","action":"details","value":"run-1842"}` and `{"type":"separator"}`.
- `BarSegment::click_value()` replaces the removed public `value` field for readers (render hit-testing, notifications, core widgets).
- `progress` is the one shape with no string payload, because its `value` is the numeric fill: it may be clickable, but the click carries no `LUVUS_MODULE_BAR_VALUE`. Documented in the Bar guide, the module guide, and the UHP reference.
- Compatibility: segments with no click payload no longer serialize `"value": null` (the key is omitted). Input with an explicit `null` still deserializes. Serializing a progress segment no longer emits `value` twice.

## Why

Fixes #306.

`BarSegment` flattened the `type`-tagged `BarSegmentKind` enum and also declared its own `value: Option<String>`. Serde drains the outer struct's declared fields from the buffered content map before the internally tagged enum sees it, so progress's numeric `value` was always claimed as a string. There was no shape that published a progress bar.

## Verification

- [x] `cargo test --locked`
- [x] Manual testing

`cargo test bar::` → 39 passed, 0 failed. Full `cargo test --locked` → 1545 passed, 4 failed; those 4 (`platform::tests::unix_stoppable_pid_accepts_a_luvus_executable_with_arguments`, two `app::session_menu::tests::stopped_session_*`, `app::tests::keyboard_copy_mode_yanks_history_and_cancel_restores_its_viewport`) fail identically on `main` under the parallel run and pass in isolation — timing-sensitive, unrelated.

New tests: `documented_progress_segment_keeps_its_numeric_value`, `progress_width_defaults_to_eight_and_stays_bounded`, `click_capable_segments_keep_their_string_value`, `segments_round_trip_through_json_without_a_duplicate_value_key`, `the_documented_content_example_parses_and_validates` (the verbatim segment array from the UHP methods docs), and `bar::render::tests::spacer_and_separator_click_values_reach_the_hit_target` — which renders a region and asserts each emitted `BarHit` carries its own segment's payload, so a dropped `spacer`/`separator` value cannot come back silently.

End to end on an isolated server (`LUVUS_HOME` in a temp dir, `examples/modules/ci-bar` linked):

```text
$ luvus bar push --id status --content '[{"type":"progress","value":23,"total":100,"width":6}]'
{"changed": true, "key": "example.ci-bar:status", "type": "ok"}

$ luvus bar list
... {"type":"progress","value":3,"total":7,"width":8,"tone":"normal","action":null} ...
... {"type":"badge","text":"2","tone":"error","action":"details","value":"run-1842"} ...

$ luvus bar push --id status --content '[{"type":"progress","value":101,"total":100}]'
invalid_request: progress requires 0 <= value <= total and total > 0
```

A progress push without `width` lists with `"width": 8`, and a badge with a 300-byte `value` still returns `action value is too long or contains controls`. Rendering verified from a real module publishing four progress segments: https://github.com/RizRiyz/luvus/pull/307#issuecomment-5566303082

## Notes

- `luvus module link <relative path>` resolves against the server's working directory, not the client's. Unrelated to this fix, but worth a docs note.

## Summary by CodeRabbit

- **New Features**
  - All supported bar segment types can carry optional click payloads.
  - Click interactions preserve each segment’s associated payload when rendered and selected.

- **Behavior Updates**
  - Progress segments use their value for numeric fill and do not provide a click payload.
  - Payloads for progress interactions should be provided through adjacent text or badge segments.

- **Documentation**
  - Updated bar interaction guidance to clarify click payload behavior across segment types.

### DIFF
--- a/src/bar/mod.rs
+++ b/src/bar/mod.rs
@@ -63,17 +63,25 @@ pub enum BarTone {
 pub enum BarSegmentKind {
     Text {
         text: String,
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        value: Option<String>,
     },
     Symbol {
         symbol: String,
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        value: Option<String>,
     },
     State {
         state: String,
         #[serde(default)]
         label: Option<String>,
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        value: Option<String>,
     },
     Badge {
         text: String,
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        value: Option<String>,
     },
     Progress {
         value: u64,
@@ -99,17 +107,17 @@ pub struct BarSegment {
     pub tone: BarTone,
     #[serde(default)]
     pub action: Option<String>,
-    #[serde(default)]
-    pub value: Option<String>,
 }
 
 impl BarSegment {
     pub fn text(text: impl Into<String>, tone: BarTone) -> Self {
         Self {
-            kind: BarSegmentKind::Text { text: text.into() },
+            kind: BarSegmentKind::Text {
+                text: text.into(),
+                value: None,
+            },
             tone,
             action: None,
-            value: None,
         }
     }
 
@@ -118,17 +126,30 @@ impl BarSegment {
             kind: BarSegmentKind::Separator,
             tone: BarTone::Muted,
             action: None,
-            value: None,
+        }
+    }
+
+    pub fn click_value(&self) -> Option<&str> {
+        match &self.kind {
+            BarSegmentKind::Text { value, .. }
+            | BarSegmentKind::Symbol { value, .. }
+            | BarSegmentKind::State { value, .. }
+            | BarSegmentKind::Badge { value, .. } => value.as_deref(),
+            BarSegmentKind::Progress { .. }
+            | BarSegmentKind::Spacer { .. }
+            | BarSegmentKind::Separator => None,
         }
     }
 
     pub fn display_width(&self) -> usize {
         match &self.kind {
-            BarSegmentKind::Text { text } | BarSegmentKind::Symbol { symbol: text } => text.width(),
+            BarSegmentKind::Text { text, .. } | BarSegmentKind::Symbol { symbol: text, .. } => {
+                text.width()
+            }
             BarSegmentKind::State { label, .. } => {
                 1 + label.as_deref().map_or(0, |label| 1 + label.width())
             }
-            BarSegmentKind::Badge { text } => text.width() + 2,
+            BarSegmentKind::Badge { text, .. } => text.width() + 2,
             BarSegmentKind::Progress { width, .. } => *width as usize,
             BarSegmentKind::Spacer { width } => *width as usize,
             BarSegmentKind::Separator => 5,
@@ -149,9 +170,9 @@ impl BarSegment {
             Ok(())
         };
         match &self.kind {
-            BarSegmentKind::Text { text } => check_text("text", text)?,
-            BarSegmentKind::Symbol { symbol } => check_text("symbol", symbol)?,
-            BarSegmentKind::State { state, label } => {
+            BarSegmentKind::Text { text, .. } => check_text("text", text)?,
+            BarSegmentKind::Symbol { symbol, .. } => check_text("symbol", symbol)?,
+            BarSegmentKind::State { state, label, .. } => {
                 if !["blocked", "working", "done", "idle", "unknown"].contains(&state.as_str()) {
                     return Err(format!("unknown state {state:?}"));
                 }
@@ -159,7 +180,7 @@ impl BarSegment {
                     check_text("state label", label)?;
                 }
             }
-            BarSegmentKind::Badge { text } => check_text("badge", text)?,
+            BarSegmentKind::Badge { text, .. } => check_text("badge", text)?,
             BarSegmentKind::Progress {
                 value,
                 total,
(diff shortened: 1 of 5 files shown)


knowsuchagency:
Verified end to end from a real module on a release build of this branch (`5b2464b`), driven by a module publishing four `progress` segments — one per provider, `value` numeric, `width` 5, toned:

four progress segments rendered in the bottom-right bar

- Push accepted on first try and stored as typed segments (`luvus bar list` → `{"type":"progress","value":47,"total":100,"width":5,"tone":"success"}`); the same payload on the 0.13.4 release is refused with `invalid type: integer 23, expected a string`, exactly as #306 describes.
- Rendering picks up `tone` (Codex above is `warning`) and the bar falls back to the widget's `compact_content` when the region is narrow, same as text segments.
- Full `cargo test --locked` on this branch: 1545 passed, 4 failed — the 4 (`platform::tests::unix_stoppable_pid_accepts_a_luvus_executable_with_arguments`, `app::session_menu::tests::stopped_session_*`, `app::tests::keyboard_copy_mode_yanks_history_and_cancel_restores_its_viewport`) fail identically on `main` under the parallel run and pass in isolation, so they are timing-sensitive and unrelated to this change.

One observation from integrating, not for this PR: the bottom-right region is capped at `MAX_BAR_REGION_WIDTH` (100 columns) and shared with `core:runtime-status`, so a module widget wider than ~66 columns is always shown compact regardless of terminal width. The docs' "256 display columns per widget" reads as if that were the practical limit; worth a note.

RizRiyz:
@knowsuchagency The progress fix and test coverage look correct. One compatibility issue should be addressed before merging.

The current Bar contract and renderer allow `action` and `value` on every segment type. With this change, `spacer` and `separator` segments can still trigger actions, but their click values are silently discarded because `click_value()` returns `None`.

Preserve the optional click value for `Spacer` and `Separator`, include both in `click_value()`, and add a regression test confirming that their values survive parsing and reach the rendered hit target.

`Progress` can remain without a string click value because its numeric `value` already represents progress. That limitation should be stated in the Bar documentation.

knowsuchagency:
Addressed in `0b18243`.

- `Spacer` and `Separator` carry an optional `value` again, and `click_value()` returns it — so `progress` is now the only shape without a string payload, since there `value` is the numeric fill. `{"type":"separator"}` and `{"type":"spacer","width":1}` still deserialize unchanged (the field is `#[serde(default, skip_serializing_if = "Option::is_none")]`, so it also does not appear in serialized output when absent).
- Regression test `bar::render::tests::spacer_and_separator_click_values_reach_the_hit_target`: parses `text`/`spacer`/`separator`, each with `action` + `value`, asserts all three survive parsing, then renders the region and asserts each emitted `BarHit` carries its own segment's payload. It fails on `5b2464b` (`every click payload survives parsing`, `[Some("run-1842"), None, None]`) and passes on `0b18243`.
- Documented the `progress` limitation in three places, since all three describe the click contract: the [Bar guide](website/src/content/docs/docs/guides/bar.mdx) segment list, the module guide's `LUVUS_MODULE_BAR_VALUE` paragraph, and the UHP reference's `ui.bar.push` section — each says a `progress` segment may be clickable but its click carries no `LUVUS_MODULE_BAR_VALUE`, and points at a neighbouring `text`/`badge` segment for the payload.

`cargo fmt`, `cargo clippy --all-targets` (clean) and `cargo test bar::` (39 passed) on the new head. No changelog entry, as `changelog/v0.13.4.md` is already released and there is no unreleased file to append to — say the word if you want one opened for the next version.

RizRiyz:
Approved!

PR #312, opened 2026-09-07
fix(input): negotiate native Windows input records

### LINKED ISSUE #225: Windows: Ctrl+V pastes as keystroke stream; image paste and hyperlinks also broken — works in Herdr, fails in Luvus
### What happened?

When I paste multi-line text with Ctrl+V into a pane inside luvus (Windows Terminal doing the paste), the content arrives truncated and the newlines are dropped — a multi-line snippet collapses into one mangled line and only part of it lands in the pane.

Expected: the full multi-line text should be forwarded to the pane's PTY as pasted.

The same paste into the same kind of pane under Herdr 0.8.2 (also running inside Windows Terminal) lands intact, and the identical Windows Terminal keybindings are in effect for both — so the divergence is inside luvus's input forwarding pipeline (client → named pipe → server PTY), which on Windows re-encodes every input record.

My Windows Terminal settings relevant to paste: `ctrl+v` → `Terminal.PasteFromClipboard`, `warning.multiLinePaste": "never"`.

### Steps to reproduce

1. Run `luvus` inside Windows Terminal on Windows.
2. Copy a multi-line snippet (e.g. a short Python script with blank lines).
3. Focus a pane (plain pwsh or an agent pane) and press Ctrl+V.
4. Pane receives a truncated, newline-collapsed version of the text. Paste the same clipboard content into a herdr pane and it arrives complete.

### Area

input

### Environment

- OS + version: Microsoft Windows 10 Home, 10.0.19045 (build 19045)
- Terminal app + version: Windows Terminal 1.24.11911.0
- Luvus version and server status:

```text
status   running
session  default
version  v0.13.2
```

```text
luvus 0.13.2

  ✓ core    the multiplexer (panes · tabs · agents) needs no external tools
  ✓ git   git version 2.45.2.windows
  ✓ gh    gh version 2.97.0 (2026-07)
  ✓ ssh   OpenSSH_for_Windows_9.5p1
  ✓ curl  curl 8.9.1 (Windows)
  ✓ logs    directory writable
  · keys    run `luvus doctor` outside a luvus pane to test your terminal

All set — you're good to go. ✓
```

### Anything else?

This looks related to the past Windows input-pipeline fixes (#61 AltGr→Ctrl chords, #43 navigation modifiers, #66 F-keys) — a paste is a large burst of input records with embedded newlines, which may be the stress case that drops/corrupts part of the burst.

A CLI-level workaround exists (`luvus pane send <id> <text>`), but interactive Ctrl+V paste is unusable for multi-line content.

RizRiyz:
## Summary

The live Windows report disproved the earlier implementation. Reading legacy `INPUT_RECORD` values directly was not enough because Luvus never enabled virtual terminal input or requested the Windows Terminal Win32 input record protocol. The decoder was therefore waiting for a stream the terminal had not been asked to send.

This revision negotiates that protocol, decodes the documented six-field key reports, reconstructs bracketed paste before IPC, and fixes the independent multiline `luvus pane send` path.

## What changed

- Enable `ENABLE_VIRTUAL_TERMINAL_INPUT` while a Windows client is attached and restore the original console mode on exit.
- Request Win32 input records with `CSI ? 9001 h` and disable the mode on detach.
- Decode the documented `CSI Vk;Sc;Uc;Kd;Cs;Rc _` reports before translating keyboard input.
- Reconstruct one bracketed paste while preserving CRLF, blank lines, UTF-16 surrogate pairs, repeats, literal control data, and backspace.
- Preserve Ctrl, Alt, Shift, AltGr, Ctrl+Shift letters, navigation keys, function keys, focus events, resize events, and mouse modifiers.
- Keep physical Escape immediate and bound incomplete control sequences and damaged pastes with explicit recovery deadlines.
- Correct both row and column offsets for native mouse records.
- Make `luvus pane send` request paste semantics so multiline CLI text uses the child pane bracketed-paste mode instead of raw typing semantics.
- Add an opt-in Windows input trace for live diagnosis. Set `LUVUS_WINDOWS_INPUT_TRACE=1` with `LUVUS_LOG=debug`. The trace is off by default. Raw UTF-16 key units can reveal typed or pasted content, so traces must be reviewed before sharing.
- Retain the Crossterm fallback when native Windows input mode cannot be established.

## Scope

This changes only the Windows host-input reader plus the CLI multiline send path. macOS and Linux keep their existing interactive input path. No dependency was added.

The record path preserves Ctrl+V and Ctrl+mouse input needed by child applications and Luvus hyperlink handling. It does not introduce a separate image clipboard transport. Live Windows verification is still required before claiming image clipboard behavior or closing #225.

## Verification

- `cargo test terminal::host_input -- --nocapture`
- `cargo test pane_send_requests_atomic_paste_semantics -- --nocapture`
- `cargo test pane_input_methods_report_rejection_and_run_is_one_action -- --nocapture`
- `cargo clippy --locked --target x86_64-pc-windows-gnu --all-targets -- -D warnings`
- `cargo test --locked`
- Isolated debug server check under `~/.luvus-dev` confirmed multiline `pane send` retained line order and content without touching the production server.

The full local suite passed with 1,615 tests, 0 failures, and 3 ignored tests. The Windows target compiles cleanly with strict Clippy. A live Windows Terminal test remains required.

## Summary by CodeRabbit

* **New Features**
  * Improved Windows terminal input handling for keyboard, mouse, focus, Unicode, modifiers, control characters, and paste events.
  * Added native Windows console input with automatic fallback when unavailable.
  * `pane send` now pastes text atomically, including multiline content.

* **Bug Fixes**
  * Incomplete paste sequences now expire instead of blocking input.
  * Improved decoding of incomplete input sequences and backspace characters during paste.
  * Invalid values for the paste option are now rejected with a clear error.

* **Documentation**
  * Updated CLI help and documentation to describe text pasting consistently.

### DIFF
--- a/src/app/dispatch.rs
+++ b/src/app/dispatch.rs
@@ -2687,8 +2687,18 @@ impl App {
             "pane.send_input" => {
                 let id = self.resolve_pane(p)?.ok_or_else(not_found)?;
                 let text = p.get("text").and_then(|v| v.as_str()).unwrap_or("");
+                let paste = match p.get("paste") {
+                    None => false,
+                    Some(Value::Bool(paste)) => *paste,
+                    Some(_) => {
+                        return Err((
+                            "invalid_request".to_string(),
+                            "paste must be a boolean".to_string(),
+                        ))
+                    }
+                };
                 let pane = self.panes.get(&id).ok_or_else(not_found)?;
-                let result = if p.get("paste").and_then(|value| value.as_bool()) == Some(true) {
+                let result = if paste {
                     pane.try_send_paste(text)
                 } else {
                     pane.try_send(text.as_bytes())
@@ -9221,6 +9231,19 @@ command = ["true"]
             panic!("expected bracketed paste")
         };
         assert_eq!(bytes, b"\x1b[200~first\nsecond\x1b[201~");
+        let error = app
+            .dispatch(
+                "pane.send_input",
+                &json!({
+                    "pane": pane.0.to_string(),
+                    "text": "must not be sent",
+                    "paste": "true",
+                }),
+            )
+            .expect_err("a non-boolean paste value must be rejected");
+        assert_eq!(error.0, "invalid_request");
+        assert_eq!(error.1, "paste must be a boolean");
+        assert!(rx.try_recv().is_err());
         drop(rx);
         for (method, params) in [
             (
--- a/src/app/dispatch.rs
+++ b/src/app/dispatch.rs
@@ -2688,8 +2688,12 @@ impl App {
                 let id = self.resolve_pane(p)?.ok_or_else(not_found)?;
                 let text = p.get("text").and_then(|v| v.as_str()).unwrap_or("");
                 let pane = self.panes.get(&id).ok_or_else(not_found)?;
-                pane.try_send(text.as_bytes())
-                    .map_err(|message| ("send_failed".to_string(), message))?;
+                let result = if p.get("paste").and_then(|value| value.as_bool()) == Some(true) {
+                    pane.try_send_paste(text)
+                } else {
+                    pane.try_send(text.as_bytes())
+                };
+                result.map_err(|message| ("send_failed".to_string(), message))?;
                 Ok(json!({"type":"ok"}))
             }
             "pane.read" => {
@@ -9199,6 +9203,24 @@ command = ["true"]
         };
         assert_eq!(bytes, b"echo hi\r");
         assert!(rx.try_recv().is_err());
+        app.panes[&pane]
+            .engine
+            .lock()
+            .unwrap()
+            .advance(b"\x1b[?2004h");
+        app.dispatch(
+            "pane.send_input",
+            &json!({
+                "pane": pane.0.to_string(),
+                "text": "first\nsecond",
+                "paste": true,
+            }),
+        )
+        .unwrap();
+        let crate::terminal::pty::InputAction::Bytes(bytes) = rx.try_recv().unwrap() else {
+            panic!("expected bracketed paste")
+        };
+        assert_eq!(bytes, b"\x1b[200~first\nsecond\x1b[201~");
         drop(rx);
         for (method, params) in [
             (
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -146,7 +146,7 @@ panes / agents:
   pane focus <id>            focus a pane (jumps to its workspace/tab)
   pane move [<id>] (--tab <n> | --new-tab)  move a pane within its workspace
   pane run [<id>] <cmd...>   run a command in a pane
-  pane send [<id>] <text>    send raw text to a pane
+  pane send [<id>] <text>    paste text into a pane
   pane read [<id>]           print a pane's recent output
   pane status [<id>]         print a pane's agent status and history metrics (any workspace)
   pane processes [<id>]      list cached executable identities without exposing arguments
@@ -3236,6 +3236,7 @@ fn parse(args: &[String]) -> Result<(String, Value)> {
             let text = tail().join(" ");
             let mut obj = serde_json::Map::new();
             obj.insert("text".to_string(), json!(text));
+            obj.insert("paste".to_string(), json!(true));
             ("pane.send_input".into(), with_pane(obj))
         }
         ("pane", "read") => ("pane.read".into(), with_pane(serde_json::Map::new())),
@@ -4463,6 +4464,28 @@ mod tests {
             let args = argv(raw);
             assert_eq!(command_help_request(&args), None, "{raw}");
         }
+    }
+
+    #[test]
+    fn pane_send_requests_atomic_paste_semantics() {
+        let (method, params) = parse(&argv("luvus pane send 9 first second")).unwrap();
+        assert_eq!(method, "pane.send_input");
+        assert_eq!(params.get("pane").and_then(Value::as_str), Some("9"));
+        assert_eq!(
+            params.get("text").and_then(Value::as_str),
+            Some("first second")
+        );
+        assert_eq!(params.get("paste").and_then(Value::as_bool), Some(true));
+
+        let args = ["luvus", "pane", "send", "9", "first\nsecond"]
+            .into_iter()
+            .map(String::from)
+            .collect::<Vec<_>>();
+        let (_, params) = parse(&args).unwrap();
+        assert_eq!(
+            params.get("text").and_then(Value::as_str),
+            Some("first\nsecond")
+        );
     }
 
     #[test]
--- a/src/i18n/cli.rs
+++ b/src/i18n/cli.rs
@@ -1278,15 +1278,15 @@ static HELP: &[Translation] = &[
         "패널에서 명령어 실행"
     ),
     tr!(
-        "send raw text to a pane",
-        "enviar texto sin procesar a un panel",
-        "enviar texto bruto para um painel",
-        "envoyer du texte brut à un volet",
-        "Rohtext an einen Bereich senden",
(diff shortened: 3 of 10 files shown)


problaems:
## Live test result: not fixed

Tested `fix/windows-native-paste` @ d5795d1 (release build, self-reports 0.13.4; client **and** server both confirmed running from that exact binary via `server status` and process inspection), Windows 10 Home 19045, Windows Terminal 1.24.11911.0. Ran it in an isolated named `--session` so my installed 0.13.2 was untouched, and verified the two were genuinely separate instances (different tab counts, separate server processes, separate sockets).

**All three symptoms from #225 are unchanged:**

1. **Ctrl+V multiline paste** — still streams in as individual keystrokes with newlines dropped/executed progressively. In an agent pane, a multiline paste still never collapses into Claude Code's `[Pasted text #1 +N lines]` placeholder (which a real bracketed paste produces). Identical to 0.13.2, and identical to the behavior I reported after #228 shipped in 0.13.4.
2. **Image paste** — still does nothing; an image on the clipboard pastes into no pane. Identical to 0.13.2. The PR contains no image-handling code at all.
3. **Ctrl+Click hyperlinks** — I did not re-test this one on this build, but nothing in the PR touches OSC 8 handling, so I have no reason to expect a change.

For completeness: the PR's automated suite passes on my machine too — 33/33 targeted `terminal::host_input` / `paste` tests, clean release build, and the PR reports the full 1,608-test suite green. That's expected and worth stating explicitly, because it's diagnostic: those tests feed the decoder **hand-crafted input records** the test author chose. They cannot observe what Windows Terminal 1.24 actually delivers during a real paste injection on Windows 10 — the record shapes, their order, their modifiers, their pacing, or even whether the new native reader engages instead of silently falling back to crossterm. Every fix attempt so far has been validated only against that synthetic reality, and every live test against the real terminal has failed identically. That gap is the whole failure mode, and it's why I asked in #225 for an opt-in input trace (`LUVUS_LOG_INPUT=1` or similar, dumping raw console records plus the decoder's emitted events to the client log). This branch still doesn't include one — I searched. The real record shapes WT sends remain unobservable, so the next fix is still guesswork. I'd happily run a debug build with that trace and report it.

## This is the second fix attempt, and the gap repeats

Some history that I think matters for evaluating this PR:

- **#228** was the first attempt: a `HostInputDecoder` reconstructing bracketed paste from crossterm events. Its own verification notes stated the author could not live-test on Windows (no MinGW toolchain in their macOS environment) — validated by unit tests and CI compile only. It was merged, shipped in 0.13.4, and my live test failed: symptom unchanged. That report is in #225 with reproduction details.
- **#312 (this PR)** is the second attempt: it bypasses crossterm entirely with a native `ReadConsoleInputW` reader and a new theory — hold the record stream until the `ESC[201~` terminator instead of letting embedded newlines act as Enter presses mid-flood. The PR description again states "A live Windows Terminal verification is still required before closing the associated report." I performed exactly that verification. It fails identically.

So we now have two consecutive merges, each validated synthetically, each failing the same live test on the same machine, same terminal, same settings — while Herdr, running in that identical environment, passes it. And the specific failure-mode candidates I listed in my #225 follow-up (coalesced INPUT_RECORDs, records with modifiers, a different functional-key encoding of `~`, release records interleaved, the prefix window being too tight) were never confirmed or refuted, because the tracing that would answer it still doesn't exist.

## What #225 actually reported, versus what this PR does

#225 was deliberately framed as **three symptoms of one root cause** — not a multiline-paste timing bug:

1. Ctrl+V multiline text pastes stream in as individual keystrokes, newlines executing progressively, no bracketed-paste placeholder.
2. **Image paste is dead** — an image on the clipboard cannot be pasted into any pane; Herdr accepts it.
3. **Ctrl+Click on OSC 8 hyperlinks printed in pane output does nothing**; Herdr makes them clickable.

The root cause framing: Windows terminal hosts (conhost, PowerShell console, Windows Terminal) deliver paste **the way a modern application does — a single paste injection** — not as an emulated typing flood. My `settings.json` (posted in full as a gist in the issue: `ctrl+v` → `Terminal.PasteFromClipboard`, `warning.multiLinePaste: "never"`, plus `sendInput` word-delete bindings) is what puts WT into that injection mode. A terminal multiplexer sitting between that host and the pane applications must pass that injection through atomically (re-wrapping as one bracketed paste when the child asks for DECSET 2004). Luvus's Windows input path fails to do that, for all three input types.

The issue also explicitly asked for a **direction**: Herdr is Luvus's base, and the correct handling already exists in the codebase this project derives from — "Multiline text paste is bracketed so shells and agent prompts receive it as one paste instead of submitting each line separately" (herdr.dev/docs/windows-beta, quoted in the issue). The request was to carry that handling over or reimplement it, not to invent a new theory of the input stream.

What this PR actually does:

- Rewrites one stage — reading console records and reconstructing bracketed paste — with a bounded `ReadConsoleInputW` batch reader and a hold-until-terminator decoder. That stage is real, carefully built, and well-tested against synthetic records. It is also **the only thing it does**.
- It does not touch image paste (symptom 2) — no clipboard-format handling, not mentioned in the body or diff.
- It does not touch hyperlinks (symptom 3) — no OSC 8 / click-routing changes, not mentioned in the body or diff.
- It does not port or reference Herdr's working Windows paste path, despite that being the requested direction and the strongest available evidence (a working reference implementation in the direct ancestor project).
- It does not model the actual paste-injection mode my settings put WT into; it models a legacy keystroke flood — the thing the issue was arguing Luvus *incorrectly* assumes.
- It does not add the input trace requested in #225, so the fix's central assumption (what records WT actually sends during a paste injection) remains unverified against reality.

In short: the issue described one root cause manifesting in three input types and pointed at a working reference implementation; the PR re-derived a narrower single-symptom theory, implemented it against synthetic input, and left two of the three symptoms and the reference implementation unaddressed. That's why I believe it missed: it solves a problem adjacent to the one reported, validated in an environment (synthetic records) that cannot falsify it.

## The controlled comparison that isolates Luvus

The Herdr comparison from the issue is worth restating because it is a controlled experiment with exactly one variable:

- Same machine, same Windows 10 19045, same Windows Terminal 1.24.11911.0 window, same `settings.json`, same clipboard content.
- Herdr 0.8.2 pane: multiline paste arrives complete and newline-intact as one bracketed paste (placeholder appears), image paste works, hyperlinks Ctrl+Click.
- Luvus pane (0.13.2, 0.13.4 post-#228, and now this PR build): all three fail, identically, across all three versions.

The terminal-side paste injection is byte-identical for both applications — they sit in the same terminal with the same settings. The divergence is inside Luvus's input forwarding (client → named pipe → server PTY), and the PR, by its own description, re-encodes that pipeline again without closing the gap to what its own base already does.

## Second, independent finding: multiline corruption in the server → pane write path

While isolating variables I found a second corruption site this PR doesn't touch — with **no keystrokes involved at all**. #225 itself suggested `luvus pane send <id> <text>` as a workaround; that path is also broken:

```
luvus pane send 5 "echo AAA\necho BBB\necho CCC"
```

Pane 5's PSReadLine buffer then contained (unexecuted, in this order):

```
echo CCC
>> echo BBB
>> echo BISECT_Becho AAA
>> echo BISECT_A
```

Lines sent as A, B, C arrived as **C, B, and A fused onto the preceding line with its newline lost** — reversed, with a line boundary dropped. Reproduced twice, consistently (a 2-line send shows the same reversal: B on the prompt line, A in the continuation). This is the server→pane write path, no client input reader involved, so even a perfect paste reconstruction would still land scrambled at the pane. It likely needs its own fix before paste behavior can be fully correct — and it means the documented CLI workaround for #225 is itself unreliable for multiline content.


Reference: https://github.com/RizRiyz/luvus/issues/225

PR #317, opened 2026-09-08
feat(input): add dedicated Windows image paste

RizRiyz:
## Summary

This adds a dedicated Windows clipboard image path while preserving the existing text input behavior.

- Read registered PNG, CF_DIBV5, and CF_DIB data only after an explicit Ctrl+V or Ctrl+Shift+V gesture
- Normalize common 24-bit and 32-bit Windows bitmaps into PNG without adding an image or compression dependency
- Carry the bounded PNG through the existing binary client transport
- Stage the image under private selected-session storage and paste its path into the originating client's focused terminal pane
- Keep text paste and unsupported clipboard content on the existing key path
- Document local and remote image paste behavior

## Safety and resource bounds

Clipboard access remains client-owned, so a Windows client attached over SSH reads the clipboard on the machine in front of the user. The server chooses the destination path and never accepts a client-provided filename.

The implementation validates PNG structure and checksums on both sides, caps payloads at 48 MiB, limits dimensions to 8192 pixels and 12 million total pixels, uses random create-new filenames, and performs bounded cleanup only during an explicit image paste. It adds no timer, watcher, polling loop, network endpoint, or new dependency.

Image paths are delivered only when a normal terminal pane owns input. Luvus modals, dashboards, native file views, copy mode, and navigation modes reject the delivery instead of leaking it into the pane underneath. The prompt is not submitted automatically.

## Verification

- `cargo fmt --all --check`
- `cargo test clipboard_image -- --nocapture`
- focused client, local-mode, protocol, and pane-routing tests
- `cargo check --target x86_64-pc-windows-gnu --tests`
- `cargo clippy --target x86_64-pc-windows-gnu --all-targets -- -D warnings`
- `cargo test --locked` with 1632 passed and 3 ignored benchmarks
- `npm run build` in `website/`
- isolated debug server lifecycle under `~/.luvus-dev` using the `image-paste-test` named session

A live Windows Terminal clipboard test was not available on this macOS development machine. The Windows implementation was cross-compiled and linted, while the platform-neutral DIB fixtures run on every host and verify row order, color channels, alpha, malformed input rejection, bounds, staging permissions, and cleanup.

## Summary by CodeRabbit

- **New Features**
  - Added Windows image pasting from the system clipboard with Ctrl+V or Ctrl+Shift+V.
  - Images are securely transferred and pasted as file paths without submitting the prompt.
  - Remote sessions support image pasting without opening additional ports.
  - Unsupported or unavailable images continue using normal paste behavior.

- **Bug Fixes**
  - Image pasting is ignored while modals, dashboards, file views, navigation, or copy mode control input.

- **Documentation**
  - Added guides covering Windows image pasting and remote-session behavior.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -103,7 +103,9 @@ windows-sys = { version = "0.61", features = [
     "Win32_System_Diagnostics_Debug",
     "Win32_System_Diagnostics_ToolHelp",
     "Win32_System_Console",
+    "Win32_System_DataExchange",
     "Win32_System_Kernel",
+    "Win32_System_Memory",
     "Win32_System_Threading",
 ] }
 
--- a/src/app/input.rs
+++ b/src/app/input.rs
@@ -674,6 +674,17 @@ impl App {
                 self.paste_into_focused_pane(&s);
                 false // goes to the pane; its echo (PtyData) renders it
             }
+            AppEvent::PasteImage(path) => {
+                // Image paths are terminal input, never modal text. Restrict
+                // delivery to the same normal focused-pane state that accepts
+                // ordinary typing so an image cannot leak through an overlay,
+                // native view, dashboard, or navigation mode.
+                if !self.focused_pane_accepts_image_paste() {
+                    return false;
+                }
+                self.paste_into_focused_pane(&path.to_string_lossy());
+                false // the pane's echo is the event that changes the frame
+            }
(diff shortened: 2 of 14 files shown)


RizRiyz:
@chuxubank @ngurajeka since you guys are on Windows, are you able to test this PR in real widnows device? let me know if you can help

PR #323, opened 2026-09-08
feat(machine): add persistent multi-machine control

RizRiyz:
## Automatic provisioning

- `machine add` and `machine enable` now discover a compatible remote binary and, when needed, install the exact matching published release automatically.
- Provisioning runs only for an explicit foreground CLI or scoped UHP mutation. Background links and reconnects remain read-only.
- The fixed embedded remote script supports macOS and Linux on x86_64 and ARM64, downloads the exact release archive, enforces time/output/file bounds, verifies the published SHA-256 digest, probes the extracted fleet protocol and version, then atomically replaces `~/.local/bin/luvus`.
- Explicit `--remote-binary` paths remain authoritative and are never replaced. Existing profiles default conservatively to unmanaged because their original path provenance is unknown.
- Revision, catalog-integrity, and duplicate-ID checks run before network or install work and repeat under the catalog lock before persistence.
- Source builds without a matching published release fail safely with guidance to install that build manually and pass its absolute path.

## Summary

- add revisioned, atomic SSH machine profiles with bounded validation, enable/disable, status, session discovery, and direct open commands
- add a persistent multiplexed fleet link for each enabled machine while a machine-aware TUI is open
- add an event-driven Machines dock and selector for desktop and mobile, with local and remote named-session switching
- add prepared, active, and suspended client surfaces so a candidate receives a complete frame before handoff and warm inactive surfaces retain no frame buffer or render work
- add opt-in UHP machine discovery and management through `uhp access --machines`, with revision-required mutations and control-scope enforcement
- publish CLI, machine workflow, remote access, UHP method, schema, and localization documentation

## Safety and lifecycle

- OpenSSH remains responsible for authentication; Luvus stores no passwords, keys, owner socket paths, or session secrets.
- Remote commands use fixed argv and a verified absolute Luvus binary path.
- The existing client path remains unchanged when the machine catalog is empty.
- Disabled profiles create no SSH child, timer, or per-machine worker.
- Enabled links use blocking reads and bounded queues rather than polling.
- Generation and channel fencing reject stale replies, and the current surface stays active until a candidate supplies a complete frame.
- Authentication or handshake failures settle in Attention instead of creating an unbounded reconnect loop.
- Removing or disabling the active profile returns the client safely to Local; remote panes remain owned by their remote server.

## Resource behavior

One isolated macOS debug sample at idle:

| Fixture | Client RSS | CPU | Client threads | Direct children |
| --- | ---: | ---: | ---: | ---: |
| Empty catalog | 9,504 KiB | 0.0% | 2 | 0 |
| One disabled profile | 9,792 KiB | 0.0% | 3 | 0 |
| One connected machine | 10,048 KiB | 0.0% | 4 | 1 |

These are single-run debug measurements, not release benchmarks. The connected case adds one sleeping reader and one SSH bridge child. The macOS release-size gate passes: executable +8.86%, compressed artifact +9.37%, both below the 10% blocking threshold. No dependency was added.

## Verification

- focused machine catalog, SSH, bridge, selector, suspended-surface, server geometry, schema, access, and gateway mutation tests
- real isolated debug lifecycle with a local SSH fixture: remote session discovery, complete-frame switch, input routing, Local return, warm-surface reuse, detach cleanup, and remote server survival
- `cargo fmt --all --check`
- `cargo check --target x86_64-pc-windows-gnu`
- `cargo check --target x86_64-unknown-linux-gnu`
- terminal backend live and failure conformance in isolated `LUVUS_HOME`
- `npm run build`

Interactive lifecycle behavior was validated on macOS. Linux and Windows received compile-boundary validation here; CI provides their platform coverage.

## Scope

The persistent transport lives for the machine-aware TUI lifetime. UHP machine methods manage and probe the same saved catalog but do not share that TUI transport. A shared owner-local coordinator and aggregate cross-machine Mission Control, finder, notification, and event projections remain separate future work.

This change adds persistent remote-machine management, remote binary provisioning, machine-aware session switching, and scoped machine controls.

Previously reported connection-reuse failures are resolved:
- **Warm channels outlive link generations:** the exercised reconnect path rejects a generation-1 warmed channel after the runtime reaches generation 2 and opens a new remote surface instead.
- **Stale channel gains current generation:** the exercised disconnect and local-handoff path retires the old remote channel before it can be reused under the replacement generation.
- **Replacement generation is assigned to a retired channel:** the exercised catalog-replacement path retires the active route before the local handoff, so returning to that machine opens a new surface rather than sending traffic to the retired channel.

Scoped machine controls also reject read-only requests to add, enable, or remove machines, while authorized control requests reach the machine API.
</details>

**What T-Rex did**
- Ran the focused generation-fencing test for warm\_remote\_surface and the federated IPC test suite; both tests passed, showing that stale Prepared/Resize reuse is not possible and the only remote path is a fresh OpenSurface.
- Executed the stale-channel generation harness to compare historical vulnerable stamping with current behavior; observed that invalidation clears the old surface and future reuses require a nonzero channel and matching generation.
- Ran the catalog replacement channel harness to validate retirement behavior; before retirement, channel 41 could switch back to Prepared/Resize, but after retirement channel 0 must OpenSurface on a new nonzero channel, and the latest regression test passed.
- Validated authorization gating for machine APIs with a scoped UHP harness; a read-only caller cannot invoke mutating machine methods due to the authorization checks.

<a href="https://app.greptile.com/trex/runs/23129525/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

<sub>Reviews (6): Last reviewed commit: ["test(settings): remove racy theme timing..."](https://github.com/rizriyz/luvus/commit/762df629654d900ffe4ecd01e9a21875e935c244) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=61580068)</sub>


yunuskorkmaz:
Good to see this. A question about scope: `prepare()` rejects a Windows remote host, but `luvus --remote <windows-host>` works today and is documented as supported. So a workflow that works now becomes unavailable through a saved profile, which reads more like a deliberate cut than a gap — hence the question.

Three places enforce it:

- `src/machine/ssh.rs` — `prepare()` rejects any probe `os` outside `linux | macos`.
- `src/machine/catalog.rs` — `validate_remote_binary()` requires an absolute POSIX path (`/`-prefixed, `[A-Za-z0-9/_-.+]`), so `C:\...\luvus.exe` cannot be recorded even when `--remote-binary` is passed explicitly.
- `src/machine/ssh.rs` — `discover_binary()` is a POSIX `sh` loop, though `--remote-binary` already bypasses it.

The transport itself looks OS-neutral from here: `ssh <dest> <binary> fleet-bridge` over stdio. And the remote guide already requires OpenSSH Server plus `luvus.exe` on the non-interactive PATH for `--remote` — the same precondition `fleet-bridge` would need.

So: is the exclusion a known blocker (argv quoting through the remote default shell, `cmd.exe` vs PowerShell as the SSH `DefaultShell`, something in the bridge framing), or is it conservative because it was not testable in this pass?

Happy to help either way. I drive a Windows host over plain SSH daily and already run a JSONL-over-stdio bridge against it in another project (`ssh <host> <exe> <args>`, stdin/stdout), so I can test a branch and report back, or open a PR for the three sites above if you would like one.

RizRiyz:
@yunuskorkmaz Windows saved-machine support is now included in this branch.

The implementation now:

- accepts shell-safe absolute Windows executable paths
- probes `luvus` from the remote non-interactive `PATH` first
- records the absolute path returned by `fleet-bridge --probe`
- supports Windows fleet links and session switching
- automatically provisions the matching x86-64 release with Windows PowerShell when no compatible binary is found
- verifies the published SHA-256 and fleet identity before installation
- installs under `%LOCALAPPDATA%\luvus\fleet\v<version>\luvus.exe`
- does not request administrator access or modify `PATH`

Windows CI now checks the PowerShell script syntax, Windows path handling, and persistent-link command construction.

For real-host validation, could you test the current PR branch with your Windows SSH host?

```sh
cargo build
./target/debug/luvus machine add windows --host your-ssh-alias
./target/debug/luvus machine show windows
./target/debug/luvus machine status windows
./target/debug/luvus machine sessions windows
./target/debug/luvus machine open windows
```

It would be useful to confirm:

1. Whether the SSH server uses `cmd.exe` or PowerShell as `DefaultShell`.
2. Detection when `luvus.exe` is already on the non-interactive `PATH`.
3. Automatic provisioning when no compatible binary is found.
4. Opening a session and sending terminal input.
5. Switching Local → Windows → Local in the TUI.
6. Reconnection after the SSH connection is interrupted.

Appreciate if you can report the command output and the Windows/OpenSSH versions if anything fails.

yunuskorkmaz:
Thanks for turning this around so fast. Results from a real Windows SSH host against branch `762df62`.

**Short version:** Windows machine control works end to end, including the Machines dock and Local ↔ Windows switching. Getting a `fleet-bridge`-capable `luvus.exe` onto the host is what fails today, and there are three bugs on that path plus one trap that makes the failure silent.

**Remote host**

- Windows 11 23H2 (10.0.22631), AMD64
- OpenSSH server 10.0.0.0, `DefaultShell` = `C:\Program Files\PowerShell\7\pwsh.exe` (PowerShell 7.6.5) — **pwsh, not cmd.exe** (your question 1)
- Released `luvus 0.13.4` on the non-interactive `PATH` at `%LOCALAPPDATA%\luvus\luvus.exe`
- Local controller: this branch built on macOS 15 arm64

---

## What works

I cross-compiled the branch to `x86_64-pc-windows-gnu` from macOS (mingw-w64 16.2.0, clean release build in 39s), copied the binary to `%LOCALAPPDATA%\luvus\fleet\v0.13.4\luvus.exe`, and passed it explicitly.

```
$ luvus machine add windows --host <user@host> --remote-binary 'C:\Users\<user>\AppData\Local\luvus\fleet\v0.13.4\luvus.exe'
{ "machine": { "id": "windows", "enabled": true, "transport": "ssh",
    "remote_binary": "C:\\Users\\<user>\\AppData\\Local\\luvus\\fleet\\v0.13.4\\luvus.exe", ... } }   # 1s

$ luvus machine status windows
{ "machine": "windows",
  "probe": { "arch": "x86_64", "os": "windows", "version": "0.13.4",
             "remote_binary": "C:\\Users\\<user>\\AppData\\Local\\luvus\\fleet\\v0.13.4\\luvus.exe" },
  "state": "online" }                                                                                # 0s

$ luvus machine sessions windows
{ "sessions": [ { "name": "default", "default": true, "running": true,
    "endpoint": { "transport": "windows_named_pipe", "address": "\\\\.\\pipe\\luvus-<id>" },
    "session_dir": "C:\\Users\\<user>\\.luvus",
    "socket_path": "C:\\Users\\<user>\\.luvus\\luvus.sock" } ] }                                      # 1s
```

Direct probe output for reference:

```json
{"arch":"x86_64","binary":"C:\\Users\\<user>\\AppData\\Local\\luvus\\fleet\\v0.13.4\\luvus.exe","os":"windows","protocol":"luvus-fleet-v1","version":"0.13.4"}
```

Path validation, case/separator handling, session discovery over the named pipe, and the probe all behaved.

Then I replaced the released `luvus.exe` on the host with the same cross-built branch binary and restarted the remote server, so both ends speak protocol 7. With that, the interactive path works:

- The **MACHINES** dock renders above `WORKSPACES` with `Local` and `windows`, both reported `online`, and the selected machine's default session shown under it.
- Selecting `windows` swaps the whole product UI to the remote server — `WORKSPACES` switches from the local workspace to the two real Windows workspaces, with their branch and `~\projects\...` paths.
- Selecting `Local` switches back.
- Restart restored workspaces, tabs and panes on the remote side as expected.

Screenshots of both states in a follow-up comment.

That covers your questions 4 and 5. I have not yet exercised reconnect after an interrupted SSH connection (question 6) or sustained input under load; I can if it is useful.

---

## Bugs on the bootstrap path

**1. `powershell.exe -Command -` silently discards the provisioning script**

`install_windows` feeds the script to `powershell.exe -NoLogo -NoProfile -NonInteractive -ExecutionPolicy Bypass -Command -`. On this host that exits **0 with empty stdout and empty stderr** — the whole `try { … } finally { … }` never runs.

```sh
# prints nothing, exit 0
printf "try {\n  'IN-TRY'\n} catch {\n  'IN-CATCH'\n} finally {\n  'IN-FINALLY'\n}\n" \
  | ssh <host> powershell.exe -NoLogo -NoProfile -NonInteractive -Command -

# prints IN-TRY / IN-FINALLY
ssh <host> powershell.exe -NoLogo -NoProfile -NonInteractive -EncodedCommand <utf16le-base64>
```

A single-line command over stdin works; a multi-line block does not. Windows PowerShell 5.1 reads `-Command -` as a stream of typed commands rather than as one script.

The empty stdout then reaches `installed_path`, so the user sees `returned an unsafe executable path`, which points at the wrong thing. `-EncodedCommand` (verified above) or a temp file plus `-File` both fix it.

**2. Automatic provisioning cannot succeed yet, and it hangs instead of failing**

`INSTALL_WINDOWS` line 63 verifies the downloaded binary with

```powershell
$ProbeText = (& $Candidate fleet-bridge --probe | Out-String).Trim()
```

No published release contains `fleet-bridge` — v0.13.4 shipped 2026-09-01, this branch is newer — so the downloaded binary opens a TUI and that call blocks with no timeout inside the script.

Everything before it worked when I ran the script by hand with `-EncodedCommand`: archive and checksum downloaded, SHA-256 verified, `Expand-Archive` produced a 13,441,536-byte `luvus.exe` under `%TEMP%\luvus-machine-<guid>`. Nothing was ever installed into `%LOCALAPPDATA%\luvus\fleet`.

The version gate cannot catch this either: `verified_probe` compares `response.version` to `CARGO_PKG_VERSION` and both sides say `0.13.4` — same version string, different capability. That resolves once a release ships; it is why the failure surfaces as a hang rather than a mismatch.

Also worth noting the `windows protocol · ConPTY` job already builds `target\debug\luvus.exe` but does not upload it, so there is nothing to fetch from a PR run. An `upload-artifact` step there would let testers bootstrap a host without cross-compiling.

**3. A remote command that never exits makes the deadline unenforceable under SSH multiplexing**

`machine add windows --host <alias>` never returned — I killed it at 7m30s with no output. `sample` on the hung process:

```
machine::cli::add (cli.rs:87)
  machine::ssh::prepare_or_provision (ssh.rs:103)
    machine::ssh::prepare (ssh.rs:40)
      machine::ssh::verified_probe (ssh.rs:60)
        machine::ssh::probe (ssh.rs:145)
          machine::ssh::run_bounded (ssh.rs:200)
            machine::ssh::run_bounded_with_input (ssh.rs:305)
              std::thread::JoinHandle::join → __ulock_wait
```

That alias sets `ControlMaster auto` / `ControlPersist 600`. The deadline fires and kills the session child, but the persistent mux master inherited the stdout/stderr pipe write ends, so the reader threads never see EOF and `join()` blocks past every timeout.

It takes both conditions. The same alias is completely fine now that the remote command exits promptly — `machine add` through it succeeded in 1s. But while the remote side hangs, `PROBE_TIMEOUT` and `PROVISION_TIMEOUT` are not enforceable at all. Dropping the readers rather than joining them after a timeout kill would bound it; `-o ControlMaster=no -o ControlPath=none` on these short fixed probes would too.

**4. An older remote Luvus hangs the probe and leaks a remote process**

Released 0.13.4 does not know `fleet-bridge`, so `ssh <host> luvus fleet-bridge --probe` starts the TUI on a channel with no tty and never exits. The local side times out; the remote process survives. Six accumulated on my host during testing:

```
31808  luvus.exe fleet-bridge --probe
30712  luvus.exe fleet-bridge --probe
28384  luvus.exe fleet-bridge --probe
14220  ...\Temp\luvus-machine-<guid>\luvus.exe fleet-bridge --probe
32464  ...\Temp\luvus-machine-<guid>\luvus.exe fleet-bridge --probe
```

An unknown subcommand exiting non-zero when stdout is not a terminal would turn this into a clean error instead of a hang plus a leak, and would also fix bug 2's hang.

**5. `machine status` reports `online` for a machine that cannot actually be selected**

Before I replaced the remote binary, `machine add` and `machine status` both succeeded and reported `state: online` — the probe only checks the fleet identity and the version string. But this PR bumps the client↔server `PROTOCOL_VERSION` from 6 to 7 and adds `SurfaceInterest`, so the released 0.13.4 server on the host could not serve the fleet client. The dock showed a green machine that did nothing when selected, with nothing in the client log.

Both sides reporting `0.13.4` is what hides it: same version string, different session protocol. Having the probe also report the session protocol version, and comparing it, would turn this into a clear error at `machine add` time instead of a machine that looks healthy and is inert.

**6. `discover_binary` is POSIX-only, so its error is misleading on Windows**

The fixed `for p in …; do … done` string is a syntax error under pwsh, so the fallback reports `Luvus was not found on remote machine` even though `luvus.exe` is on the non-interactive `PATH`. Worth noting it is only reachable via bug 4: once the host had a `fleet-bridge`-capable binary, the bare-command PATH probe in `prepare()` found it with no `--remote-binary` and no fallback, in well under a second.

---

Happy to re-run against a branch with fixes, and I will report back on the interactive switching once I have driven it. I can send a PR for 1, 3 and 6 if you would rather not carry them — say which you want to own.

Local selected — the local session's workspaces:
shot-local

`windows` selected — the same client, now showing the remote Windows
server's two workspaces:
shot-windows

RizRiyz:
This PR wontfix, so I moved to new PR with more simple and clear implementation. 

@yunuskorkmaz when you have time please test this PR on windows #333

PR #325, opened 2026-09-08
feat(agent): add native Devin CLI support

paulork:
Devin CLI panes appear in the AGENTS sidebar with live state, and a pane whose exact session id Luvus recorded reopens with `devin --resume <id>`.

## What

- Registers `devin` as a built-in agent with the Cursor-shaped session contract: `discovery: None`, restore from an exact binding that was reported to Luvus and persisted (`luvus pane report --agent devin --session <id>`), and no fork. Nothing is discovered, so `luvus agent resume <id>` cannot find Devin sessions; the descriptor doc, the guides, the reference, and the Luvus skill doc say so, in the same terms Kilo Code and OpenCode 2 use.
- Treats `devin` as an ambiguous identity, since it is also a given name. Only deliberate process, launch-command, or title evidence names the agent; prose on screen does not.
- Adds two Devin-specific screen rules:
  - **blocked** on the first-run workspace-trust menu, matching the shared stem `trust the authors of` together with its `No, exit` label. Devin words that screen both as "…authors of this directory?" and "…authors of `<dir>`?", so the stem covers both without matching either label alone.
  - **working** on `esc again to interrupt` and `esc twice to interrupt`. Neither phrase contains a generic `WORKING_HINT`, so without this rule a generating Devin pane only read as working while a spinner frame happened to lead a line.
- Adds Devin to `screen_uses_non_empty_rows`, beside Claude, Codex, and Hermes. Devin paints the trust menu on the first rows of the grid and leaves the rest blank, so the ordinary 14-row bottom window sees only empty rows and the pane read as idle while waiting on the prompt. A VT-backed test renders the captured screen, asserts it sits above the bottom window, and checks the non-empty path reports blocked — the same shape as the existing Claude approval-panel test.
- Stops `filter_launch_flags` at `--` for Devin. ORCH launches Devin as `devin -- <briefing>`, and with **Settings → General → resume launch flags** on, a restored pane would otherwise replay that briefing after `devin --resume <id>`. The options before the separator still return.
- Guards the shared Claude/Copilot hook script against Devin. Devin imports Claude-format hooks from the global and project `.claude/settings.json` and marks every hook process it spawns with `DEVIN_PROJECT_DIR`; the script now exits before reporting when it is set, so a Devin pane can never be reported, displayed, or persisted as Claude (or Copilot).
- Devin's permission menu (`Yes, allow once` / `Yes, allow for this session`) already matches the generic `yes, allow` prompt. A test pins that so it is not duplicated.
- Uses Devin's documented `devin -- <prompt>` syntax for ORCH workers.
- Adds `devin` to the shared `~/.agents/skills` host, which Devin reads, keyed on the `devin` command and `$XDG_CONFIG_HOME/devin`.
- Publishes the detection-only manifest `website/public/manifests/devin.toml` (identity only, like Kilo Code's) for older Luvus versions that fetch `luvus.dev/manifests`.
- Updates the README support table, the homepage agent grid, the FAQ agent list, the agent guide and reference, the automation guide, the agent-messaging skill destinations, the shared-skill destination lists, the ORCH task-command table, `usage.rs`, and the Luvus skill doc.

## Why

Devin CLI had no native adapter. A detection manifest could name the process, but not the `--` prompt contract, the resume command, the trust screen, or the interrupt hint, so a pane running Devin stayed idle in the sidebar both while generating and while waiting on the trust prompt.

Session discovery is deliberately absent. Devin keeps history in a private SQLite store under its config root whose schema is on migration 16 and still moving, and #248 established that Luvus does not open agent databases. A workspace-to-session mapping needs a documented file or a Devin-side hook, not a database read.

## Verification

- [ ] `cargo test --locked` — see details

Details:

- `cargo fmt --all --check` — clean.
- `cargo clippy --all-targets -- -D warnings` — clean for this change. Pre-existing findings are reported in `src/agent/opencode/config.rs` and `src/terminal/host_input/console.rs`, which this PR does not touch; both files are byte-identical to `main`, and they appear to be newer lints than the CI toolchain that last passed there.
- `cargo test agent::` / `detect::` / `skill::` / `usage::` / `agent_task_commands_follow_each_cli_prompt_contract` / `automation_capabilities` / `devin_restores_its_exact_binding_without_replaying_the_briefing` — 148 passed. `cargo test integration::` — 34 passed, 1 failed: `installing_replaces_only_the_legacy_managed_hook`, which fails identically on unmodified `main` in this Windows checkout (CRLF in a fixture). The hook-script execution test is `cfg(unix)` and runs on the Linux and macOS jobs.
- `cargo test --locked` (full suite) does not complete on the Windows checkout used here: sixteen tests stall under parallel execution, and the list is byte-for-byte identical on unmodified `main` in the same checkout, so it is not introduced by this change. The Linux job ran the full suite green on the first round.
- `npm run build` in `website/` — 49 pages.
- Command, path, and screen contracts were checked against the installed `devin 3000.6.7`: `--help` for `--resume` and the `--` prompt separator, `devin skills paths` for the shared skills directory, and the trust-screen and interrupt strings taken from the shipped binary. The four screenshots below are the manual test, taken with a build of `main` and a build of this branch from the same base commit.
- Devin's hook behavior was measured, not assumed: a dump hook in a project-scope `.claude/settings.json` recorded the environment and payload of every event Devin emits (details under Notes).

## Screenshots

Same directory (`C:\luvus-devin-demo`, outside any directory Devin already trusts), same pane, same command, on both builds.

**Before — `main`, Devin waiting on the trust prompt: `no active agents`**

Before: Luvus on main, Devin waiting on its trust prompt, sidebar shows no active agents

**After — this branch, the same screen: `blocked devin`**

After: same trust screen on this branch, sidebar shows blocked devin

**After — trust accepted, Devin at its prompt: `idle devin`**

After: Devin at its prompt after trust was accepted, sidebar shows idle devin

**After — generating, "esc twice to interrupt" on screen: `working devin`**

After: Devin generating with esc twice to interrupt on screen, sidebar shows working devin

## Notes

- **Automation is declared `None` on purpose.** Devin does expose the pieces an unattended profile needs (`-p` print mode and `--permission-mode`), but its trust gate has to be resolved as well: print mode fails in an untrusted directory because it cannot show the prompt. Declaring an access level that has not been run end to end could strand a scheduled task or over-permit it, so automation is left to its own reviewed change rather than guessed here.
- **The integration hook is a separate follow-up.** Devin reads hook definitions in the Claude Code format from the `hooks` key of its own config (`~/.config/devin/config.json`, `%APPDATA%\devin\config.json` on Windows) and, as measured for this PR, from the global and project `.claude/settings.json` as well, covering `SessionStart`, `SessionEnd`, `PermissionRequest`, `PreToolUse`, `PostToolUse`, `UserPromptSubmit`, `Stop`, and `PostCompaction`. A `luvus integration install devin` would add exact live session ownership and put Devin in **Settings → Integrations**, but it is its own reviewable surface (config editing, uninstall preservation, Windows shell behavior) and is intentionally out of scope here.
- **What Devin hands a hook, measured on 3000.6.7.** Beyond the parent environment it sets exactly `DEVIN_PROJECT_DIR`, `CHISEL_SESSION_DB` (its session database), and `CLAUDE_PROJECT_DIR` — the last one mimics Claude Code, so it cannot tell the two apart; the guard keys on `DEVIN_PROJECT_DIR`, the mirror of that Claude variable. The payload carries `hook_event_name` and `session_id` (a `word-word` slug) and no `transcript_path`. Devin's hook parser is strict: an event name it does not know (`Notification`, which the Luvus Claude hook registers) rejects the whole settings file, so on this version the Luvus Claude hook is not imported at all today. The guard is there for the day that parser loosens, or for an older install that registered `SessionStart` alone.
- Windows process command lines arrive quoted from the PEB, which is how an install path with spaces (`C:\Users\Ada Lovelace\...\devin.exe`) stays one token. The identity test covers that form; no other agent test in the tree did.

Adds native Devin CLI support for exact-session restoration, terminal state detection, hook isolation, shared skills, and documentation. Executed checks confirmed that session restoration excludes the original task briefing, top-positioned trust prompts are classified correctly, and Devin-hosted hooks do not report Claude session ownership.
</details>

**What T-Rex did**
- Ran the focused session-contract harness with exact-ID resume, capability, and persisted-launch tests; both harness modes exited cleanly and all four tests passed; a permission-mode auto pane was opened to fix the login bug and Devin resumed as specified, remaining unavailable for discovery or forking.
- Executed the focused Devin terminal classification suite in Chromium with an 80×30 terminal fixture; all five tests passed, and the ordinary extraction showed a blank top portion while non-empty rows included the trust prompt and labeled it Blocked; interrupt hints were Working, the composer Idle, and prose about Devin identity did not claim Devin.
- Ran the shared hook validation with a luvus recording and a Claude SessionStart payload; in normal Claude mode a Claude session was reported, while the Devin-marked path produced no report, and the focused Rust hook runtime test also passed, confirming the guard against Devin-hosted Claude hooks.
- Validated harness stability by confirming both harness modes exited with status 0 and all four tests passed; the authored harness source and two before/after captures were saved for review.
- Observed exact terminal classification behaviors: trust content in rows 1–5 with rows 6–30 blank is absent from detection\_text(14) but retained by detection\_text\_non\_empty(14) to yield Blocked; certain prompts produced Working and Idle classifications, and prose indicating Devin did not create Devin identity avoided a false positive.

<a href="https://app.greptile.com/trex/runs/23466505/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

<sub>Reviews (3): Last reviewed commit: ["docs(skill): use full luvus commands in ..."](https://github.com/rizriyz/luvus/commit/15c00bdd56802ebc7b755e1089e955c449f60e8c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=61796070)</sub>

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,7 @@ Keyboard → Keyboard Shortcuts → Input Sources** to free `Ctrl+Space`.
 | Fx | ✓ | ✓ | No |
 | Cursor | ✓ | resume command | No |
 | Kilo Code | ✓ | exact-ID resume | No |
+| Devin | ✓ | exact-ID resume | No |
 | Gemini · Aider · Amp · Droid · Qwen · Kiro | ✓ | No | No |
 
 Live status needs no agent integration. See the
--- a/plugins/luvus/skills/luvus/SKILL.md
+++ b/plugins/luvus/skills/luvus/SKILL.md
@@ -458,9 +458,9 @@ surface:
 - OpenCode 2 Preview is a separate `opencode2` agent. Do not install the
   OpenCode V1 integration for it or infer session IDs from its live database.
 - Devin has native detection and exact-ID resume only. Do not infer session
-  IDs from its private database; `agent.resume` cannot find Devin sessions,
-  so bind a pane with `pane report --agent devin --session <id>` when the
-  exact id is known.
+  IDs from its private database; `luvus agent resume <id>` cannot find Devin
+  sessions, so bind a pane with `luvus pane report --agent devin --session
+  <id>` when the exact id is known.
 - For Hermes, `luvus integration install hermes` adds exact per-pane session
   ownership for restart resume. Detection remains native, but Luvus does not
   scan Hermes's private history store.
--- a/plugins/luvus/skills/luvus/SKILL.md
+++ b/plugins/luvus/skills/luvus/SKILL.md
@@ -457,6 +457,10 @@ surface:
   session ownership and structured usage. Without it, usage stays unavailable.
 - OpenCode 2 Preview is a separate `opencode2` agent. Do not install the
   OpenCode V1 integration for it or infer session IDs from its live database.
+- Devin has native detection and exact-ID resume only. Do not infer session
+  IDs from its private database; `agent.resume` cannot find Devin sessions,
+  so bind a pane with `pane report --agent devin --session <id>` when the
+  exact id is known.
 - For Hermes, `luvus integration install hermes` adds exact per-pane session
   ownership for restart resume. Detection remains native, but Luvus does not
   scan Hermes's private history store.
--- a/skills/luvus/SKILL.md
+++ b/skills/luvus/SKILL.md
@@ -458,9 +458,9 @@ surface:
 - OpenCode 2 Preview is a separate `opencode2` agent. Do not install the
   OpenCode V1 integration for it or infer session IDs from its live database.
 - Devin has native detection and exact-ID resume only. Do not infer session
-  IDs from its private database; `agent.resume` cannot find Devin sessions,
-  so bind a pane with `pane report --agent devin --session <id>` when the
-  exact id is known.
+  IDs from its private database; `luvus agent resume <id>` cannot find Devin
+  sessions, so bind a pane with `luvus pane report --agent devin --session
+  <id>` when the exact id is known.
 - For Hermes, `luvus integration install hermes` adds exact per-pane session
   ownership for restart resume. Detection remains native, but Luvus does not
   scan Hermes's private history store.
--- a/skills/luvus/SKILL.md
+++ b/skills/luvus/SKILL.md
@@ -457,6 +457,10 @@ surface:
   session ownership and structured usage. Without it, usage stays unavailable.
 - OpenCode 2 Preview is a separate `opencode2` agent. Do not install the
   OpenCode V1 integration for it or infer session IDs from its live database.
+- Devin has native detection and exact-ID resume only. Do not infer session
+  IDs from its private database; `agent.resume` cannot find Devin sessions,
+  so bind a pane with `pane report --agent devin --session <id>` when the
+  exact id is known.
 - For Hermes, `luvus integration install hermes` adds exact per-pane session
   ownership for restart resume. Detection remains native, but Luvus does not
   scan Hermes's private history store.
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -169,6 +169,12 @@ fn filter_launch_flags(agent: &str, launch: &[String]) -> Vec<String> {
             }
             continue;
         }
+        // Devin reads everything after `--` as the initial prompt. Replaying it
+        // on `devin --resume <id>` would run the captured task briefing again,
+        // so the separator ends the copy; the flags before it still apply.
+        if agent == "devin" && t == "--" {
+            break;
+        }
         if t.contains('=') && TAKES_VALUE.contains(&head) {
             i += 1; // glued form, e.g. --resume=<id>
             continue;
@@ -698,6 +704,25 @@ mod tests {
             f("copilot", &["--resume=old", "--banner"]),
             vec!["--banner"]
         );
+        // Devin: the flags before `--` are kept; the separator and the task
+        // briefing after it are not, and a stale selector before it still goes.
+        assert_eq!(
+            f(
+                "devin",
+                &[
+                    "--permission-mode",
+                    "auto",
+                    "--",
+                    "fix",
+                    "the",
+                    "login",
+                    "bug"
+                ]
+            ),
+            vec!["--permission-mode", "auto"]
+        );
+        assert!(f("devin", &["--", "fix", "the", "login", "bug"]).is_empty());
+        assert!(f("devin", &["--resume", "old", "--", "fix"]).is_empty());
         // Standalone selectors, a fork flag, and one-shot print mode all go.
         assert_eq!(
             f(
@@ -775,6 +800,24 @@ mod tests {
         assert_eq!(
             resume_command_with_flags("opencode2", "ses_2", &opencode2_launch).as_deref(),
             Some("opencode2 --session 'ses_2' '--standalone'\r")
+        );
+
+        // Devin keeps its option but never the `-- <briefing>` it was launched with.
+        let devin_launch = [
+            "--permission-mode",
+            "auto",
+            "--",
+            "fix",
+            "the",
+            "login",
+            "bug",
+        ]
+        .iter()
+        .map(|value| value.to_string())
+        .collect::<Vec<_>>();
+        assert_eq!(
+            resume_command_with_flags("devin", "quiet-meadow", &devin_launch).as_deref(),
+            Some("devin --resume 'quiet-meadow' '--permission-mode' 'auto'\r")
         );
 
         // All-filtered input and empty input both fall back to the plain command.
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -22,6 +22,7 @@ pub(crate) mod claude;
 pub(crate) mod codex;
 pub(crate) mod copilot;
 pub(crate) mod cursor;
+pub(crate) mod devin;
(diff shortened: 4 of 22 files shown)


RizRiyz:
@paulork The adapter structure and detection coverage are solid, but these issues need to be resolved before merge:

1. Prevent prompt replay during restore

Devin tasks launch as `devin -- <briefing>`. When launch flag restoration is enabled, Luvus currently preserves `-- <briefing>` and can append it to `devin --resume <id>`, causing the original task to run again.

Update `filter_launch_flags` so Devin stops copying arguments at `--`. Add a regression test confirming that flags before the separator are preserved while the separator and briefing are removed.

2. Align the resume claim with the actual public behavior

The adapter declares `discovery: None`, but `luvus agent resume <id>` only accepts sessions found through discovery. The command therefore returns `not_found` for Devin IDs.

Either remove the claim that `luvus agent resume <id>` works and document resume as limited to an exact binding already reported and persisted by Luvus, or implement a validated discovery or resume path that includes enough information to resolve the agent and working directory safely.

3. Prevent the Claude integration from claiming Devin panes

Devin loads hooks from `~/.claude/settings.json` by default. The existing Luvus Claude hook reports `--agent claude`, and integration reports override native detection. Users with the Claude integration installed can therefore have Devin displayed and persisted as Claude.

Add a reliable guard so the Claude hook does not report Devin sessions, with tests confirming that normal Claude reporting still works.

4. Complete public presentation parity

Add Devin to the homepage supported-agent grid and the supported-agent list in the FAQ.

Please include focused regressions for prompt replay, exact-binding restoration, and Claude-hook isolation.

RizRiyz (changes requested):
https://github.com/RizRiyz/luvus/pull/325#issuecomment-5595286549

paulork:
Thanks for the review. All four points are addressed in fff520b; the PR description is updated to match.

**1. Prompt replay** — `filter_launch_flags` now stops at `--` for Devin. Regressions: `["--permission-mode", "auto", "--", "fix", "the", "login", "bug"]` keeps only `--permission-mode auto`; a stale `--resume old` before the separator still goes; `resume_command_with_flags` yields `devin --resume 'quiet-meadow' '--permission-mode' 'auto'`. There is also an end-to-end restore test in `app::tests`: a Devin pane with a persisted exact binding snapshots `agent_launch` with the briefing in it and restores without it, with the setting on and off.

Worth knowing while you are here: the same replay reaches every agent whose launch carries a prompt flag — `--prompt` for Kilo, Kimi and OpenCode 2, `--prompt-interactive` for Gemini and Qwen, `--oneshot` for Hermes, `-p` for Antigravity — because `filter_launch_flags` only strips session selectors. I kept this PR to Devin; happy to send a descriptor-driven fix (`task_prompt_args` plus its value) as a separate PR if you want one.

**2. Resume claim** — removed. The descriptor doc, guides, reference, and the Luvus skill doc now say resume is limited to an exact binding reported and persisted by Luvus (`luvus pane report --agent devin --session <id>`), and that `luvus agent resume <id>` cannot find Devin sessions because nothing is discovered — the same terms Kilo Code and OpenCode 2 use.

**3. Claude hook isolation** — I measured what Devin 3000.6.7 does rather than guess, with a dump hook in a project-scope `.claude/settings.json`. It imports Claude-format hooks from both the global and the project file (in `-p` mode too), and adds exactly three variables to a hook's environment: `DEVIN_PROJECT_DIR`, `CHISEL_SESSION_DB`, and `CLAUDE_PROJECT_DIR` — the last one mimics Claude Code, so it cannot discriminate. Its payload has no `transcript_path`. The shared hook script now exits before reporting when `DEVIN_PROJECT_DIR` is set; Copilot's script gets the same guard since they share the template.

Tests: the installed Claude and Copilot scripts carry the guard and still report as themselves, and a `cfg(unix)` test runs the script for real against a stand-in `luvus` — a Claude `SessionStart` payload produces `pane report --agent claude --session abc-123`, the same payload under `DEVIN_PROJECT_DIR` produces nothing.

One more finding: Devin's hook parser is strict. An event name it does not know — `Notification`, which the Luvus Claude hook registers — rejects the whole settings file, so on 3000.6.7 the Luvus Claude hook is not imported at all today; the guard covers the day that parser loosens, or an older install that registered `SessionStart` alone. If you would also like a server-side check (reject a report whose agent contradicts a positively identified process tree), say so and I will add it.

**4. Presentation parity** — homepage grid and FAQ, plus the surfaces #313 and #318 touched that I had missed: the published `manifests/devin.toml` and index, the agent-messaging skill destinations, the automation guide, `usage.rs`, and the Luvus skill doc.

RizRiyz (approved):
Approved! thanks @paulork for contributing to luvus

PR #333, opened 2026-09-09
feat(machine): add persistent SSH multi-machines

RizRiyz:
Manage persistent SSH machines and their remote workspaces alongside local projects in the native workspace sidebar.

## What

- Add saved machine profiles shared across local named sessions, with CLI management and opt-in scoped UHP access.
- Connect each machine to one selected remote session and project its workspaces without creating a server per workspace.
- Preserve native workspace navigation, context menus, keyboard focus, and the named-session dropdown.
- Keep both sidebar layouts client-owned across local, remote, and named-session switches.
- Add a Remote Machine setup tab with saved profiles, connection states, Tab/Shift+Tab navigation, and retained form drafts.
- Discover compatible remote binaries and support permission-based installation. Every installation requires fresh foreground approval; reconnects never install or restart servers.
- Add bounded reconnects, compatibility checks, Windows-specific SSH/provisioning support, and a step-by-step public guide.

## Why

Remote projects should remain accessible from the same workspace navigation without replacing local session controls or duplicating remote servers. Machine profiles persist independently of individual local sessions while panes and agents remain owned by their remote server.

## Verification

- Focused saved-machine and picker navigation regression tests passed.
- All 14 federated client tests passed.
- `cargo check --tests --locked` passed.
- `cargo clippy --all-targets --locked -- -D warnings` passed.
- `cargo fmt --all --check` and `git diff --check` passed.
- `cargo build --locked` passed.
- Astro website build passed.
- Isolated macOS-to-Linux live testing covered remote workspace opening, sidebar layout switching, saved-profile reuse without duplicates, and form preservation across tab switches.

## Notes

- Native Windows SSH lifecycle testing remains outstanding; compilation and focused CI coverage do not establish runtime parity.
- The full locked test suite and release build were not rerun for the final picker/documentation changes.
- Native client protocol compatibility is required on both endpoints. Existing remote servers are never restarted automatically.
- Disabling or removing a profile disconnects client access but does not stop remote servers, panes, or agents.
- No production servers were modified for the final verification. Local test artifacts and ignored planning documents are excluded.

This change adds saved remote-machine profiles, durable preparation records, catalog revision fencing, platform-aware provisioning, and bounded connection recovery.

Earlier reported failures are no longer present: interrupted or conflicted remote setup retains recoverable preparation data; stale concurrent CLI updates are rejected; legacy session data no longer blocks catalog updates; normal Windows profile paths are accepted; repeated handshake timeouts stop at an actionable terminal state; recovery guidance conditions receipt instructions on an installation having occurred; and the FreeBSD checkout uses an immutable revision.
</details>

**What T-Rex did**
- Ran a focused post-install crash test for the planned installation and verified the test passes after plan("/opt/luvus"), then inspected the machine list to confirm no planned binary was saved for the failed installation.
- Executed the focused catalog-conflict recovery test, observed that the installation marker remained and pending() held a receipt with installed\_binary: /opt/luvus, while the real machine-list output showed the record under preparations with profile and binary details.
- Reproduced concurrent catalog mutations with two real rename commands; one commit occurred at revision 2 while the other failed with a revision conflict, and the deterministic shared-snapshot fence test passed.
- Ran handshake\_timeout tests; the focused test passed and demonstrated three expired connecting deadlines, with the third deadline reaching terminal Attention instead of scheduling another reconnect.
- Ran a focused FreeBSD checkout pin assertion; the assertion passed and reported the expected commit SHA for the pin in the current workflow.

<a href="https://app.greptile.com/trex/runs/23837993/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

<sub>Reviews (12): Last reviewed commit: ["fix(machine): harden remote recovery bou..."](https://github.com/rizriyz/luvus/commit/461db55b6d7ed5c96d193b00c229b6d4c5028b12) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=62131693)</sub>


RizRiyz:
Appreciate if you can test this @yunuskorkmaz

yunuskorkmaz:
## macOS client → Windows machine test

The PR notes say "Native Windows SSH lifecycle testing remains outstanding" — that is exactly what we covered here. Both ends run a release build of this branch (macOS aarch64 ↔ Windows 10 x86_64, OpenSSH with key auth). Testing was done both interactively by hand in the TUI and through scripted runs.

### Results against the PR's goals

**Saved machine profiles + CLI management — works.**
`machine add / list / show / status / sessions` behave as documented. The profile lives in `~/.luvus/machines.json`, revision tracking works, and `machine status` runs one bounded probe and reports `online`.

**Compatible remote binary discovery — works.**
Without `--remote-binary` and without granting `--install`, the remote binary was discovered and validated on its own:

```
"remote_binary": "C:\\Users\\...\\AppData\\Local\\luvus\\luvus.exe"
"probe": { "arch": "x86_64", "os": "windows", "version": "0.13.4" }
```

**One remote session per machine, no server per workspace — works.**
Opening the client started the Windows server over the SSH bridge with exactly one child process per machine (`ssh -T … remote-client-bridge`). The Windows named-pipe transport is reported correctly. Remote workspaces are listed under `Win PC` without spawning additional session servers.

**Client-owned sidebar / endpoint-owned content — works.**
Switching to the remote surface kept the WORKSPACES sidebar in place while the AGENTS dock switched to the agents running on Windows.

**Local ↔ remote switching — works.**
Across a Remote → Local → Remote round trip, `hostname` returned `DESKTOP-7B368A1` / `<mac>.local` / `DESKTOP-7B368A1`. Scrollback was preserved on return, and the remote server and its panes stayed alive. Remote pane geometry is correct: in a 160x45 terminal with a 25-column sidebar the remote pane reported `132x42`.

**Scoped UHP access — works.**
`uhp access --machines` returned a descriptor with `"features": {"machines": true}`. With read-only authority, `machine.list/get/status/sessions` succeeded and `machine.add` was rejected with `forbidden`. The SSH destination and the absolute binary path are omitted from the projections.

**Tests — pass.**
`cargo test machine::` passes 29/29 natively on Windows and 30/30 on macOS.

### Findings

**1. The `freebsd amd64 test` job added by this PR is red.** 1711 passed, 21 failed, including this PR's own `machine::provision::tests::provision_script_verifies_and_installs_in_the_user_prefix` and `local_build_script_streams_and_verifies_before_installing`.

**2. `--remote` reports a raw decoding error on a protocol mismatch.** This PR moves the native client protocol from 6 to 13, but both sides still report version "0.13.4", so the mismatch is not obvious. `--remote` resolves the remote binary from `PATH` (`src/main.rs:792`), so when the remote host still has an older Luvus the user sees:

```
Error: connection failed before the Luvus handshake: UnexpectedEnd { additional: 91 }
```

The `machine` path handles this properly by comparing protocols through `remote-client-info --json` first; `--remote` has no such check, and no way to point at a specific remote binary the way `machine --remote-binary` does. Since the protocol jumps seven versions, anyone using `--remote` will hit this during the upgrade.

### Setup note

The Windows machine had no MSVC linker (no Visual Studio installed), so building from source required the `stable-x86_64-pc-windows-gnu` toolchain plus mingw. Windows provisioning in this PR targets the published x86-64 MSVC release, so this is outside its scope, but Windows users building from source will hit the same wall.

### Not covered

The `+ → Remote Machine` form (Tab navigation, retained form drafts), the `--install` provisioning approval flow, reconnect/backoff and `attention` states, switching named sessions on a machine, remote work surviving `disable`/`remove`, UHP `--control` and `if_revision` conflict rejection, Linux remote hosts, and the mobile surface.

PR #382, opened 2026-09-16
fix(input): move the text cursor inside rename dialogs

### LINKED ISSUE #381: Rename dialog cannot move the text cursor
**What happened?**

In the rename dialog the text cursor is stuck at the end of the name. Left and Right do nothing, and Home, End and Delete do nothing either. So I can only add or remove characters at the end.

If I want to fix a typo in the middle of a name, I have to delete everything after it and type it all again. With a long workspace name that is annoying.

It happens in all three rename dialogs: workspace rename (right click a workspace in the sidebar), tab rename (prefix then `,`), and pane rename.

I expected the arrow keys to move the cursor inside the name, typing to insert at the cursor, Backspace to delete the character before it and Delete the one after it.

**Steps to reproduce**

1. Right click a workspace in the sidebar and choose Rename.
2. The dialog opens with the current name, for example `Recall bot issue`.
3. Press Left a few times. The cursor bar stays at the end of the name.
4. Type a letter. It lands at the end, not where you wanted it.
5. Press Home, End or Delete. Nothing happens.

**Area**

input, workspaces

**Environment**

- OS: macOS 26.4 (build 25E246)
- Terminal: Ghostty 1.3.1
- Luvus version and server status: binary v0.13.4, running server v0.12.0

```text
\   /   Luvus server
 \_/    status   running
(o_o)   session  default
/|_|\   version  v0.12.0
        socket   /Users/<user>/.luvus/luvus-client.sock
  note: this binary is v0.13.4 - run `luvus server restart` to load it
```

```text
luvus 0.13.4

  ✓ core    the multiplexer (panes · tabs · agents) needs no external tools

  ✓ git   git version 2.55.0          git tab · worktrees
  ✓ gh    gh version 2.97.0 (nixpkgs  GitHub PRs & issues
  ✓ ssh   OpenSSH_10.5p1, OpenSSL 3.  luvus --remote
  ✓ curl  curl 8.7.1 (x86_64-apple-d  luvus module search

  · logs    /Users/<user>/.luvus/logs
  ✓ logs    directory writable

  ! keys    Shift+Enter isn't distinguishable here · optional
           ↳ Luvus still works; only the modified-Enter shortcut is affected
             use Alt/Option+Enter or a terminal with the keyboard protocol

All set - you're good to go. ✓
```

**Anything else?**

This is not a version problem. The code on main (7e70fcb) behaves the same way.

`handle_ws_rename_key`, `handle_tab_rename_key` and `handle_pane_rename_key` in `src/app/mod.rs` only push a typed char onto the buffer and pop the last char on Backspace, so there is no cursor position stored anywhere. `draw_rename` in `src/ui/picker.rs` always paints the cursor bar after the whole buffer.

I have a local branch with a fix (a shared single line editor used by the three dialogs, with arrows, Home, End, Backspace and Delete, and the cursor drawn at the right spot). I will open a PR for it.

yosephbernandus:
The rename dialogs let you move the text cursor and edit anywhere in the name.

## What

- The tab, workspace, and pane rename dialogs now handle `←` / `→`, `Home`, `End`, and `Delete`. Typing inserts at the cursor, Backspace removes the character before it, Delete the one after it.
- The cursor bar is drawn where the edit lands instead of always after the whole name.
- The three dialogs share one small single-line editor in `src/app/line_edit.rs` instead of each handling keys on their own.
- The pane dialog keeps its addressable-name rule (a letter first, then letters, digits, `_` or `-`). Every edit is checked against the finished name, so a delete that would leave a digit at the front is refused the same way a bad keystroke always was. Names still lowercase as you type and the length caps are unchanged.
- `Enter` and `Esc` are untouched, and so is the name that gets saved.

## Why

Closes #381.

The three handlers in `src/app/mod.rs` only pushed a typed char onto the buffer and popped the last one on Backspace, so there was no cursor position at all. Fixing a typo in the middle of a workspace name meant deleting everything after it and typing it again. `draw_rename` in `src/ui/picker.rs` painted the cursor bar after the buffer, which matched that behavior.

## Verification

- [x] `cargo test --locked`
- [x] Manual testing on macOS 26.4 with Ghostty 1.3.1, using a debug build

`cargo test --locked` ran the whole suite: 1913 passed, 3 failed. None of the three touch this change. Two pass on a rerun (`diff_api_status_scan_is_off_loop_and_cached_lists_answer_immediately` and `observed_prompt_exited_terminal_releases_ownership_before_pane_removal`, both timing based). `files_show_hidden_defaults_on_and_the_setting_persists` also fails on a clean checkout of `main`, so it is already broken there.

New tests: three unit tests for the editor (caret moves, multi byte names, rejected edits and a stale caret), one workspace rename test that walks the reported case, and one pane rename test for the name rule.

## Screenshots

Screen recording of the workspace rename dialog before and after the change:

## Notes

Only tested on macOS. The keys involved are plain `Left`, `Right`, `Home`, `End`, `Delete` and `Backspace`, so no platform specific input path is touched.

The docs page `website/src/content/docs/docs/reference/keybindings.mdx` gets a short paragraph about these keys.

<details><summary>Greptile Summary</summary>

Adds cursor-aware editing to tab, workspace, and pane rename dialogs. The shared editor now supports caret navigation, insertion, Backspace, and Delete; rename modals render the caret position; coverage and keybinding documentation were added.
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(input): move the text cursor inside ..."](https://github.com/rizriyz/luvus/commit/f91a1324643c9f67ba53d440013077874413728f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=64847563)</sub>


RizRiyz (approved):
Approved! thanks @yosephbernandus

